PR #16, opened 2021-01-22
[Phoe Chuan Yi, Benny] iP

Bennyphoe:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,23 @@
+import java.time.LocalDate;
+import java.time.LocalTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Date;
+
+public class Deadline extends Task{
+
+
+
```
cheeerynne: Should these empty lines be removed?

src/main/java/Duke.java
```diff
@@ -1,10 +1,34 @@
...
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+
+        new Duke("./data/tasks.txt").run();
+
+
     }
+
+
```
cheeerynne: Should these empty lines be removed? I noticed the same issue in a few other places too.

src/main/java/Event.java
```diff
@@ -0,0 +1,24 @@
+import java.time.LocalDate;
+import java.time.LocalTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Date;
+
+public class Event extends Task{
```
cheeerynne: Perhaps you can use Egyptian style brackets here!

src/main/java/Parser.java
```diff
@@ -0,0 +1,124 @@
...
+        this.taskList = taskList;
+        this.storage = storage;
+    }
+
+    /**
+     * Returns a boolean if the user inputs "bye"
+     * @param userInput takes in user input as a String
+     * @return a boolean to indicate whether user wants to quit the application
+     */
+    public boolean determineAction(String userInput) {
```
cheeerynne: Would this method be clearer to the readers if the method name is named to sound like a boolean? Perhaps a suggestion like 'isExit' would be more suitable?

src/main/java/Parser.java
```diff
@@ -0,0 +1,124 @@
...
+    public Parser(TaskList taskList, Storage storage) {
+        this.taskList = taskList;
+        this.storage = storage;
+    }
+
+    /**
+     * Returns a boolean if the user inputs "bye"
+     * @param userInput takes in user input as a String
+     * @return a boolean to indicate whether user wants to quit the application
+     */
```
cheeerynne: Perhaps you can add a full stop behind each parameter description to comply with the Javadoc comments form! Same for the other Javadoc comments.

src/main/java/Storage.java
```diff
@@ -0,0 +1,164 @@
+import java.io.*;
```
cheeerynne: Would it be better if the imported classes are listed explicitly?

cheeerynne (commented):
Code generally looks good and majority of your code follows the coding standard! 
Variables and methods are named clearly, making it easy to read and follow. 
Perhaps a few slight changes to comply with the coding standard would be good.

src/test/TaskListTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+
+    @Test
+    void addTask_newToDo_success() {
+        TaskList taskList = new TaskList(testList);
+        taskList.addTask("todo borrow book");
+        assertEquals(1, taskList.getNewStorage().size());
+    }
+
+    /*
```
litone01, line 26: Will it be better to remove codes commented out when pushed to master? If necessary, can store them in another branch:)

src/main/java/mainclasses/Ui.java
```diff
@@ -0,0 +1,41 @@
...
+            parser.checkInput(userInput);
+            output = parser.determineAction(userInput);
+        } catch (DukeException ex) {
+            System.out.println(ex.getMessage());
+        }
+        return output;
+    }
+
+    /**
+     * Initialise the first dialog from Duke
```
litone01, line 34: Should it be "Intialises" rather than "Initialise"? A full stop "." is missing at the end of sentence.

src/main/java/mainclasses/TaskList.java
```diff
@@ -0,0 +1,126 @@
...
+        this.newStorage = initialisedStorage;
+    }
+
+
+    /**
+     * Prints the task in a list format, includes description, task type and whether task is completed
+     * @param num int index of task in the list
+     * @param task the task to be printed out
+     */
+    public String printTask(int num, Task task) {
```
litone01, line 26: Should the name of the method be "getTask" or "getTaskInfo"? This is because this method is not "printing" the task information and it is more to return the task information before it gets printed.

src/main/java/mainclasses/TaskList.java
```diff
@@ -0,0 +1,126 @@
...
+
+    public List<Task> getNewStorage() {
+        return newStorage;
+    }
+
+    /**
+     * Finds Tasks which description contains a certain keyword
+     * @param keyWord This is the keyword used to filter the list of tasks
+     */
+    public String findTasksWithKeyword(String keyWord) {
```
litone01, line 127: Very minor point here. Will it be better to change "keyWord" to "keyword" as it is a single word?

src/main/java/mainclasses/Task.java
```diff
@@ -0,0 +1,70 @@
...
+
+    protected String getOnlyDescription() {
+        return this.description;
+    }
+
+    public boolean getDone() {
+        return this.done;
+    }
+
+    public void setDone(boolean set) {
```
litone01, line 35: Should "set" be changed to "isDone" since it is a boolean?

src/main/java/mainclasses/Task.java
```diff
@@ -0,0 +1,70 @@
+package mainclasses;
+
+import java.time.LocalDate;
+import java.time.LocalTime;
+
+
+public class Task {
+    protected String description;
+    protected boolean done;
```
litone01, line 9: Should "done" be changed to "isDone" since it is a boolean?

src/main/java/mainclasses/Deadline.java
```diff
@@ -0,0 +1,25 @@
+package mainclasses;
```
litone01, line 1: Should the pakage name be more specific? For example, those about UI can be undert the package ui and those about storage can be about storage.

litone01 (commented):
Great job and nice to look through your work:)

PR #18, opened 2021-01-22
[Kumaran] iP

kumsssss:


src/main/java/AddCommand.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    @Override
+    public boolean isBye() {
+        return info[0].equals("bye");
+    }
+
+
+    /**
+     * adds a task to the list using the given information.
```
YuFeng0930: Should JavaDoc start with an upper case letter? (Adds a task...)

src/main/java/TaskList.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+
+    /**
+     * creates a Task object and adds it to the TaskList object
+     *
+     * @param info String array containing details of the user input
+     * @throws DukeException if the user input is not specified properly
+     */
+    void add(String[] info) throws DukeException {
```
YuFeng0930: Maybe should rename it as "addTask"? Besides, this function seems a bit too long.

src/main/java/Ui.java
```diff
@@ -0,0 +1,115 @@
...
+        return "OOPS! The file hasn't been created yet";
+    }
+
+
+    /**
+     * Prints lines to show the user what task has been added.
+     *
+     * @param tasks TaskList of all the tasks
+     */
+    String addedTask(TaskList tasks) {
```
YuFeng0930, line 41: Most of the functions in this class act as a printer. Hence, maybe these functions should be renamed as e.g. "printAddedTask"? Otherwise, it's not easy to know "addedTask" (and other functions in this class) are actually printers.

src/main/java/Ui.java
```diff
@@ -0,0 +1,115 @@
...
+        String result = "Nice! I've marked this task as done:" + "\n" + "    " + task.toString();
+        return result;
+        //System.out.println("Nice! I've marked this task as done:");
+        //System.out.println("    " + task);
+    }
+
+    String printFilteredList(TaskList tasks) {
+        ArrayList<Task> list = tasks.getList();
+        int listLength = tasks.getListLength();
+        String result = "Here are the matching tasks in your list:";
```
YuFeng0930, line 98: Constant names must be all uppercase. --> RESULT? There are also other constant names need to be change in your code.

YuFeng0930 (commented):
Most of the things are doing quite well. Only some simple coding standard violations need to take notes.

PR #19, opened 2021-01-22
[Yu Jiaxiang] iP

litone01:


src/main/java/duke/ui/UI.java
```diff
@@ -0,0 +1,89 @@
+package duke.ui;
+
+import java.util.Scanner;
+
+/**
+ * Interacts with user by receiving input and printing results from user commands.
+ */
+public class UI {
```
guanyz: Should this be in PascalCase (Ui)?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,69 @@
...
+     */
+    public static void main(String[] args) {
+        new Duke().run(args);
+    }
+
+    private void run(String[] args) {
+        start();
+        runUntilExit();
+        exit();
+    }
```
guanyz, line 38: I like how clean this method is!

src/main/java/duke/tasklist/TaskList.java
```diff
@@ -0,0 +1,152 @@
...
+        };
+        return res;
+    }
+
+    /**
+     * Lists the tasks in the task list
+     * @return the information of the tasks present
+     */
+    public static String[] getAllTaskListInfo() {
+        String[] res;
```
guanyz: Perhaps this variable could be given a clearer name? It took me a bit to realize what it was meant to store.

guanyz (commented):
Nice implementation of Duke! The code looks nice and clean overall.

src/main/java/duke/Duke.java
litone01: Thanks!

src/main/java/duke/tasklist/TaskList.java
litone01: Yep sure! I think it is a valid point!

src/main/java/duke/ui/UI.java
litone01: Thanks for the point! I think I will check with the coding standard again to see if it should be Ui or UI. Personally find it more natural to use UI hahaha.

PR #31, opened 2021-01-22
[Lai Yu Heem] iP

yuheem:


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,44 @@
+package duke;
+
+import duke.command.Command;
+import duke.task.TaskList;
+
+public class Duke {
+    private final Storage storage;
+    private TaskList tasks;
+    private final Ui ui;
```
onnwards: ### Should `storage` and `ui` be in uppercase? I noticed this in other areas as well.

Honestly, I wouldn't consider this to be violations, but I'll leave it to you to judge:
- Your variables are final whenever possible, which is good,
- However, that means it probably should be written in uppercase according to the coding convention.
- My personal belief is that only static final variables should be written in uppercase, but the coding convention isn't clear on that.
- However, an example they provided was `final int COLOR_RED = 1;` which leans towards the former,

### I would leave them as non-final to avoid this iffy-ness.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public Duke(String pathname) {
+        ui = new Ui();
+        storage = new Storage(pathname);
```
onnwards: should pathname be pathName instead?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,86 @@
...
+
+public class Storage {
+    private File file;
+
+    public Storage(String pathname) {
+        this.file = new File(pathname);
+    }
+
+    public List<Task> loadTasksFromFile() throws DukeException {
+        List<Task> taskList = new ArrayList<>();
```
onnwards: should taskList be written with a plural word instead? ie. tasks

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,22 @@
...
+public class DeleteCommand extends Command {
+    public DeleteCommand(String arguments) {
+        super(arguments);
+    }
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        int index = Integer.parseInt(this.arguments);
+        Task deletedTask = tasks.get(index);
+        tasks.remove(index);
```
onnwards: not a coding standard violation, but `List.remove()` returns a `Task`, so you can actually do Task deletedTask = tasks.remove(index) instead :)

src/main/java/duke/command/CommandType.java
```diff
@@ -0,0 +1,35 @@
...
+    DELETE("delete");
+
+    private final String label;
+    private static final Map<String, CommandType> BY_LABEL = new HashMap<>();
+
+    static {
+        for (CommandType command : values()) {
+            BY_LABEL.put(command.label, command);
+        }
+    }
```
onnwards: This is the first time i'm seeing this syntax. How does this work?

onnwards (commented):
LGTM, just a few minor coding standard violations. Saw some interesting syntax in your code, am interested to know how it works :)

Other things I noticed:
Will `"data/tasks.txt"` have cross platform issues? (Windows uses `\`)

src/main/java/duke/command/CommandType.java
yuheem: It is called a static block. Whenever a class is loaded into memory Java will execute the block only once.

src/main/java/duke/Duke.java
yuheem: Great point. I will probably leave it as non-final as well.

src/main/java/duke/Duke.java
yuheem: Pathname is a single word so I chose to use pathname instead of pathName. Here is Java's own documentation for their File class for reference: [https://docs.oracle.com/javase/8/docs/api/java/io/File.html](https://docs.oracle.com/javase/8/docs/api/java/io/File.html )

src/main/java/duke/Storage.java
yuheem: Good catch!

PR #32, opened 2021-01-22
[Ernest Pan] iP

ampan98:


src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,20 @@
...
+    private final String message;
+
+    /**
+     * Creates an exception using a message.
+     *
+     * @param message The message to return in toString method.
+     */
+    public DukeException(String message) {
+        super(message);
+        this.message = message;
```
tlylt: Hi, I suppose you could have done without `super(message);` and it will still work as intended? 
Let me know if there is actually an intended effect of your `super(message);` 😃

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,126 @@
...
+    public static String process(String input, TaskList tasks) throws DukeException {
+        if (input.equals("bye")) {
+            return null;
+        }
+
+        StringBuilder output = new StringBuilder();
+        if (input.equals("list")) {
+            output.append("  Tasks in list:\n");
+            for (int i = 0; i < tasks.size(); i++) {
+                output.append("  ").append(i + 1).append(".").append(tasks.get(i)).append("\n");
```
tlylt: Perhaps the use of `String.format()` could have been better in terms of understandability?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,165 @@
...
+
+        scene = new Scene(mainLayout);
+        stage.setScene(scene);
+        stage.show();
+
+        // Arrange nodes for GUI
+        stage.setTitle("Duke");
+        stage.setResizable(false);
+        stage.setMinHeight(600.0);
+        stage.setMinWidth(400.0);
```
tlylt: Perhaps extracting these magic numbers out as named constants will be easier to understand?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * Executes operations on the task list based on given input and
+     * returns the response by the program.
+     *
+     * @param input Input from user.
+     * @param tasks List of tasks.
+     * @return Program response as a String.
+     * @throws DukeException If input is not recognised.
+     */
+    public static String process(String input, TaskList tasks) throws DukeException {
```
tlylt, line 18: Overall, the `process` method seems to be longer than desired. Quoting our textbook:
> Be wary when a method is longer than the computer screen, and take corrective action when it goes beyond 30 LOC (lines of code). The bigger the haystack, the harder it is to find a needle.

Perhaps extract out some smaller methods?

tlylt (commented):
Good job on progressing to JavaFx that fast. 
Just some comments for your considerations, please.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,126 @@
...
+
+            try {
+                tasks.markAsDone(inputNumber - 1);
+            } catch (IndexOutOfBoundsException ex) {
+                throw new DukeException("  That number is invalid.");
+            } catch (IOException ex) {
+                throw new DukeException("  Unable to save tasks.");
+            }
+
+            output.append("  Successfully marked as done:\n");
```
hengyiqun: Perhaps you could try exploring the use of "\t" to create tabs, instead of coding the spaces out.

src/main/java/duke/Parser.java
hengyiqun: I agree with the above. While it is possible to understand currently, the use of `String.format()` may improve understandability further.

src/main/java/duke/DukeException.java
ampan98: Yeah its redundant, thanks for noticing. Will remove.

src/main/java/duke/Ui.java
ampan98: Thanks, will do so.

src/main/java/duke/Parser.java
ampan98: Alright thanks. Will do this as part of code quality improvements.

src/main/java/duke/Parser.java
ampan98: Understood, thanks for the suggestion. Will do this as part of code quality improvements.

PR #35, opened 2021-01-22
[Dylan Ghee Shao Kang] iP

zatkiller:


src/main/java/duke/DialogBox.java
```diff
@@ -0,0 +1,61 @@
...
+import javafx.collections.FXCollections;
+import javafx.collections.ObservableList;
+import javafx.fxml.FXML;
+import javafx.fxml.FXMLLoader;
+import javafx.geometry.Pos;
+import javafx.scene.Node;
+import javafx.scene.control.Label;
+import javafx.scene.image.Image;
+import javafx.scene.image.ImageView;
+import javafx.scene.layout.HBox;
```
BenedictBCJJ, line 18: It might be good to separate the imports based on packages so that there is consistency and is readable  (coding standard)

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,164 @@
...
+import javafx.scene.image.Image;
+import javafx.scene.layout.AnchorPane;
+import javafx.scene.layout.VBox;
+import javafx.stage.Stage;
+import duke.exceptions.DukeException;
+import duke.exceptions.UnknownInputException;
+import duke.tasks.DeadlineTask;
+import duke.tasks.EventTask;
+import duke.tasks.Task;
+import duke.tasks.ToDoTask;
```
BenedictBCJJ: Same thing about import statements, consistent and readable. (coding standard)

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,81 @@
...
+            throw new DukeOnlyIoException();
+        }
+    }
+
+    /**
+     * Returns a arraylist of String objects that will be used to load the TaskList
+     *
+     * @return An ArrayList of String
+     */
+    public ArrayList<String> load() {
```
BenedictBCJJ: I think having a more verbose method name would help in identifying what it does exectly

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * UI class that is responsible for reading user input and printing
+ */
+public class Ui {
+    /**
+     * Constructor to initialize the UI Class
+     */
+    public Ui() {
+    }
```
BenedictBCJJ: If the class is not storing any variables, it might not be needed to initiate an instance of it, instead, you can just make all the methods static

BenedictBCJJ (commented):
Overall very well done, just that import statements seem inconsistent between longer and shorter statements. Good job.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.ArrayList;
+import java.util.Scanner;
+
+import duke.exceptions.DukeOnlyIoException;
+
+/**
+ * Storage class that is used for the reading and writing of the list file
+ */
+public class Storage {
+
```
Assyarul: Stating class variables as null seems redundant as removing " = null " gives the same effect. Not really a coding violation but lesser code is written as a result.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,280 @@
...
+        this.inputList.add(task);
+    }
+
+    /**
+     * Returns The task inside the task list at the index specified
+     *
+     * @param index index of task inside the tasklist
+     * @return A task
+     */
+    public Task get(int index) {
```
Assyarul: Perhaps it is better to have a longer method name here e.g getTask ?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,280 @@
...
+
+        return task;
+    }
+
+    /**
+     * Returns The String that will be used to write into the file.
+     *
+     * @return A string to be written into the file.
+     */
+    public String getListToWrite() {
```
Assyarul: getListToString seems more appropriate here as you are returning a String as the output.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,129 @@
...
+    
+    /**
+     * Prints the welcome message.
+     */
+    public static String printWelcome() {
```
Assyarul: The methods here should be renamed to reflect the actual use of the function. No printing is done within the method. e.g getWelcomeMessage(), getByeMessage()

Assyarul (commented):
Looks very clear in terms of structure. The naming of methods can be improved to show users their use cases.

src/main/java/duke/DialogBox.java
zatkiller: I tried doing that but checkstyle throws an "Extra separation in import group" error

src/main/java/duke/Duke.java
zatkiller: Same issue as above, I tried to separate the imports previously for better readability but it didn't manage to pass checkstyle

src/main/java/duke/Storage.java
zatkiller: Thanks for pointing this out, will make the changes

src/main/java/duke/Storage.java
zatkiller: Good point even though this is not really a coding violation

src/main/java/duke/TaskList.java
zatkiller: Removed the method as it is no longer in use

src/main/java/duke/TaskList.java
zatkiller: Thanks, getListToString does sound nicer :D

src/main/java/duke/Ui.java
zatkiller: Noted, will change to avoid confusion

PR #39, opened 2021-01-23
[Tan Jun Wei] iP

w2vgd:


src/main/java/duke/MainWindow.java
```diff
@@ -0,0 +1,129 @@
+package duke;
+
+import duke.commands.*;
```
tjtanjin: It might be better to import the classes individually instead of doing a wildcard import :)

src/main/java/duke/MainWindow.java
```diff
@@ -0,0 +1,129 @@
...
+                + "|____/ \\__,_|_|\\_\\___|\n";
+        // Create a new Label control
+        String welcomeMsg = "Hello! I'm Duke.\nWhat can I do for you?";
+        dialogContainer.getChildren().add(DialogBox.getDukeDialog(welcomeMsg, dukeImage));
+
+        primaryStage.setTitle("Duke");
+        primaryStage.setMinWidth(400.0);
+        primaryStage.setMinHeight(500.0);
+        
+    }
```
tjtanjin: Since this isn't a constructor, setter or getter, perhaps there should be a header comment for this method as well?

src/main/java/duke/commands/CommandResult.java
```diff
@@ -0,0 +1,76 @@
...
+
+import duke.tasks.TaskList;
+
+/**
+ * Represents the result of a command after execution.
+ */
+public class CommandResult {
+    private final String messageForUser;
+    private final TaskList updatedTaskList;
+    private final boolean isExiting;
```
tjtanjin: It might be best to capitalise the names of constants even if they are private attributes. I noticed this in several other files too so it might be good to do a quick check on this :) For your convenience, the bottom of this [quick guide](https://github.com/nus-cs2103-AY2021S2/forum/issues/93) shares how the `checkstyle.xml` can be updated to catch these violations :)

src/main/java/duke/ui/DialogBox.java
```diff
@@ -0,0 +1,53 @@
...
+import javafx.fxml.FXMLLoader;
+import javafx.geometry.Pos;
+import javafx.scene.Node;
+import javafx.scene.control.Label;
+import javafx.scene.image.Image;
+import javafx.scene.image.ImageView;
+import javafx.scene.layout.HBox;
+
+import java.io.IOException;
+import java.util.Collections;
```
tjtanjin: Should the import statements for java packages come before the import statements for third party packages?

tjtanjin (commented):
LGTM, almost. Just a few nits to fix, great job :D

src/main/java/duke/MainWindow.java
w2vgd: Thanks for pointing that out! After configuring gradle, it seems that my intellij settings have resetted 😭

src/main/java/duke/MainWindow.java
w2vgd: Noted. Will add one in the next commit.

src/main/java/duke/commands/CommandResult.java
w2vgd: Thanks for the link to the guide 😄 
Do you think capitalizing the instance attributes or removing the `final` modifier will be more appropriate in this case?

src/main/java/duke/ui/DialogBox.java
w2vgd: Overlooked this 😓 Thanks for pointing this out.

src/main/java/duke/commands/CommandResult.java
tjtanjin: Unless the application is behaving incorrectly with the final keyword, I am inclined to think that capitalizing the instance attributes would be the better option. The final keyword reliably ensures that the state of the attribute is never modified and this seems especially important for the taskList, which should not have a new list object unintentionally overwriting it. I have to admit I am not too experienced with java and have only read up on the use of final recently so I am open to and would appreciate any advice on this as well :D

w2vgd: Shall ask the forum 😃

PR #43, opened 2021-01-23
[He Qiyuan] iP

QY-H00:


src/main/java/duke/Momo.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+
+    /**
+     * Get responses from Momo using the input of users
+     *
+     * @param input input of user
+     * @return answer of Momo
+     */
+    public String getResponse(String input) {
```
ssoonwee: Do you think is it possible to break this code into 2? Makes it easier to troubleshoot and for future reuse. :)

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,150 @@
...
+     * Returns a Deadline by parsing the input behind the command.
+     * Used when the command is deadline.
+     * Works when the input is "{description} /by YYYY-MM-DD".
+     *
+     * @param input input of users.
+     * @return a task of Deadline corresponding to the input.
+     * @throws ParseException if the description is empty.
+     * @throws DateTimeParseException if the format of data time is not correct.
+     */
+    private static Deadline parseDeadline(String input) throws ParseException, DateTimeParseException {
```
ssoonwee, line 101: I see the `parseDeadline` and `parseEvent` has quite a number of similarities, do you think can these 2 be placed under one method instead? 🤔

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,26 @@
+package duke.task;
+
+import java.time.LocalDate;
+
+/**
+ * One kind of Task that need to be done before the date.
```
ssoonwee, line 6: Maybe instead of 'One' kind can be 'A'? Just my thoughts :D 
need -> needs

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,69 @@
...
+        this.filePath = Paths.get(path);
+    }
+
+    /**
+     * Loads the task list file from filePath.
+     *
+     * @return a list of task corresponding to the file.
+     * @throws IOException if such filePath does not exist.
+     */
+    public List<Task> load() throws IOException {
```
ssoonwee: I feel the load method could be named more specific such as defining it as `loadTasks`, in case if the Storage method has multiple load commands in the future. Just my thoughts :)

ssoonwee (commented):
Overall, I really love the structure of your code. There is a clear distinction between the classes and methods, and they are quite easy to read. I have left a few pointers on your code on the potential improvement. If possible, do let me know how you feel about them! Good job so far! 👍

src/main/java/duke/Parser.java
QY-H00: Sure. I think it will be more consice to place them in one method. Thank you!

PR #44, opened 2021-01-23
[Ng Boon Hong] iP

NBH99:


src/test/java/DukeTest.java
```diff
@@ -0,0 +1,48 @@
...
+public class DukeTest {
+
+    @Test
+    public void parserTest1() throws Exception {
+        assertEquals("borrow book",
+                Parser.getTask("borrow book"));
+    }
+
+    @Test
+    public void parserTest2() throws Exception {
```
SiTingST: Perhaps you can change parserTest2/3/4 method name to something more meaningful?

src/main/java/TaskList.java
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+    }
+
+    /**
+     * Search for tasks which take place at that timing.
+     *
+     * @param time string to be searched.
+     */
+    public void dateTask(String time) {
```
SiTingST: Would be better to avoid using misleading names?

src/main/java/Ui.java
```diff
@@ -0,0 +1,64 @@
...
+    public void showLine() {
+        System.out.println("-------------------------------" +
+                "--------------------------------");
+    }
+
+    public void showList() {
+        System.out.println("Here are the tasks in your list:");
+    }
+
+    public void showAdd(Task t, int size) {
```
SiTingST: Perhaps can change it to showAddedMessage?

src/main/java/DoneCommand.java
```diff
@@ -0,0 +1,34 @@
+public class DoneCommand extends Command {
+
+    public DoneCommand(String action, String info) {
+        super(action, info);
+    }
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException {
+        int num;
+        Task t;
```
SiTingST: Might be a good idea to rename t ?

src/test/java/DukeTest.java
NBH99: Thanks for the input.

src/main/java/TaskList.java
NBH99: Thanks for the input.

src/main/java/Ui.java
NBH99: Thanks for the input.

src/main/java/DoneCommand.java
NBH99: Thanks for the input.

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public AddCommand(String action, String info) {
+        super(action, info);
+    }
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException {
+        String task, time;
+        int size;
+        Task t;
```
fairyinabottle4: I think you can select a better variable name than t, as the scope here is for the whole method.

src/main/java/Duke.java
```diff
@@ -1,10 +1,45 @@
...
+    public void run() {
+        ui.showWelcome();
+        boolean isExit = false;
+        while (!isExit) {
+            try {
+                String fullCommand = ui.readCommand();
+                ui.showLine();
+                Command c = Parser.parse(fullCommand);
+                c.execute(tasks, ui, storage);
+                isExit = c.isExit();
```
fairyinabottle4: I think it is confusing to have the variable name the same as the method that is called. I suggest that you change the variable name(see above comment)

src/main/java/Duke.java
```diff
@@ -1,10 +1,45 @@
...
+    }
+
+    /**
+     * Main body of Duke program.
+     * Command is created by parsing the input and then executed.
+     * Error message is shown if there is DukeException.
+     */
+    public void run() {
+        ui.showWelcome();
+        boolean isExit = false;
```
fairyinabottle4: While isExit is a valid variable name, I think you can choose a better one like shouldExit or canExit. From an English readability point of view, it will help things slightly.

src/main/java/ListCommand.java
```diff
@@ -0,0 +1,21 @@
+public class ListCommand extends Command {
```
fairyinabottle4, line 1: I like that you managed to encapsulate all your commands under a single Command class

src/main/java/Parser.java
```diff
@@ -0,0 +1,124 @@
...
+     *
+     * @param string
+     * @return
+     * @throws DukeException
+     */
+    public static Command parse(String string) throws DukeException {
+        String action, info;
+        Command command;
+
+        if (string.contains(" ")){
```
fairyinabottle4: There should be a space between the ) and the {

fairyinabottle4 (commented):
Overall, I think your code adheres to the coding standard very well. There are only a few issues with the naming of some variables and methods

src/main/java/AddCommand.java
NBH99: Thanks for the input.

src/main/java/Parser.java
NBH99: Thanks for the input.

src/main/java/ListCommand.java
NBH99: Thanks for the input.

src/main/java/Duke.java
NBH99: Thanks for the input.

src/main/java/Duke.java
NBH99: Thanks for the input.

PR #45, opened 2021-01-23
[Pavitra] ip

pavz02:


src/main/java/Parser.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Parses the inputs given by the user.
+ */
+public class Parser {
+    /**
+     * Determines the keyword and processes the input based on the keyword.
+     */
+    static String parse(String inp) {
+        String[] spl = inp.split(" ", 2);
```
VisnuRavi: Could perhaps use a actual word to name the variable spl, as it is being used throughout the parse code

VisnuRavi: also for the checkSplLength below

src/test/java/UiTest.java
```diff
@@ -0,0 +1,26 @@
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class UiTest {
+    private static final String SPACE = "     ";
+    private static final String LINE = "\n     <<>><<>><<>><<>><<>><<>><<>><<>><<>><<>><<>><<>><<>><<>><<>><<>><<>><<>><<>><<>><<>><<>><<>><<>><<>><<>>\n";
+
+    @Test
+    public void CorrectTaskFormat_TypesOfTasks_FormatOfTask() {
```
VisnuRavi: I believe the starting letter should be lower caps according to coding standard

VisnuRavi (commented):
I think your code is great! I have mentioned just a few things that I would change, but other than that should be good to merge!

src/main/java/Parser.java
pavz02: Thank you for your review :)

PR #51, opened 2021-01-23
[Zhou Yi Kelvin] IP

zhoube:


kieron560 (commented):
Overall looks good in the Coding Standards! Although one nitpick about your javadocs is that there is supposed to be punctuation behind each parameter description as per the Java Coding Standards.

zhoube:
Thank you!

src/main/java/Kelbot.java
```diff
@@ -4,22 +4,32 @@
     public static void main(String[] args) {
         Scanner sc = new Scanner(System.in);
         System.out.println("Hello! I'm Kelbot\n" + "What can I do for you?");
-        List<String> toDoList = new ArrayList<>();
+        List<Task> toDoList = new ArrayList<>();
         String input = sc.nextLine();
+        String[] commands = input.split(" ");
         while (true) {
```
prerthan99: instead of while(true) I believe we can also do for(;;)

src/main/java/Kelbot.java
```diff
@@ -4,22 +4,32 @@
...
             if (input.equals("bye")) {
                 break;
             } else if (input.equals("list")) {
-                // List out all the to-dos
+                // List out all the tasks
                 for (int i = 1; i <= toDoList.size(); i++) {
-                    System.out.println(i + ". " + toDoList.get(i - 1));
+                    System.out.println(i + "." + toDoList.get(i - 1));
                 }
+            } else if (commands[0].equals("done")) {
```
prerthan99: perhaps make a variable for commands[0] = firstCommand for ease of reading

zhoube: Ah okay thanks!

src/main/java/kelbot/DeadlineTask.java
```diff
@@ -0,0 +1,23 @@
+package kelbot;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.FormatStyle;
+
+public class DeadlineTask extends Task {
```
gycc7253, line 7: Doesn't public class need header documentation also?

src/main/java/kelbot/DialogBox.java
```diff
@@ -0,0 +1,56 @@
...
+import javafx.fxml.FXMLLoader;
+import javafx.geometry.Pos;
+import javafx.scene.Node;
+import javafx.scene.control.Label;
+import javafx.scene.image.Image;
+import javafx.scene.image.ImageView;
+import javafx.scene.layout.HBox;
+
+
+public class DialogBox extends HBox {
```
gycc7253, line 16: same as above for documentation in my opinion

src/main/java/kelbot/Kelbot.java
```diff
@@ -0,0 +1,115 @@
...
+        ui = new UI();
+        storage = new Storage(path);
+        try {
+            taskList = storage.load();
+        } catch (KelbotException e) {
+            System.out.println(e.getMessage());
+            taskList = new TaskList();
+        }
+    }
+    public String getResponse(String input) {
```
gycc7253, line 33: this method needs documentation too haha

src/main/java/kelbot/Kelbot.java
```diff
@@ -0,0 +1,115 @@
...
+                response = "Invalid Date!";
+            }
+        } catch (ArrayIndexOutOfBoundsException e) {
+            response = "Which task are you referring to?";
+        } catch (DateTimeParseException e) {
+            response = "Date cannot be empty!";
+        }
+        storage.save(taskList.getTaskList());
+        return response;
+    }
```
gycc7253: this deep nested loop can be simplified perhaps? but i actually have something similar myself haha, maybe i should look into it too...

src/main/java/kelbot/Storage.java
```diff
@@ -0,0 +1,61 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.util.ArrayList;
+
+public class Storage {
+    private java.nio.file.Path path;
+    private boolean fileExists;
```
gycc7253: do you think this name should be doesFileExisits maybe?

src/main/java/kelbot/Storage.java
```diff
@@ -0,0 +1,61 @@
...
+            } catch (Exception ex) {
+                ex.printStackTrace();
+            }
+        } else {
+            throw new KelbotException("No existing data file detected. A new file will be created");
+        }
+        return null;
+    }
+    /**
+     * Save Task List as data.txt file
```
gycc7253: I remember the start of method documentation should be third-person singular? like Saves Task list as ....

src/main/java/kelbot/EventTask.java
```diff
@@ -0,0 +1,23 @@
...
+     * @param date The date that the Event Task must be completed by
+     */
+    public EventTask(String name, LocalDate date) {
+        super(name);
+        this.date = date;
+    }
+    @Override
+    public String toString() {
+        return "[E] " + super.toString() + " at " + date.format(DateTimeFormatter.ofLocalizedDate(FormatStyle.LONG));
+    }
```
gycc7253, line 24: this method needs comment also i think

src/main/java/kelbot/Parser.java
```diff
@@ -0,0 +1,93 @@
...
+                    break;
+                }
+                taskName += " " + commands[i];
+            }
+            date = LocalDate.parse(eventDate);
+        } else {
+            isValid = false;
+        }
+    }
+    public Command getCommand() {
```
gycc7253, line 84: All these public methods not documented sia... are u planning to document them later maybe?

gycc7253 (commented):
Nice reviewing your code haha, very nice job! Looks good to me!

src/main/java/kelbot/Parser.java
zhoube: I believe prof has mentioned that trivial methods do not require documentation!

PR #53, opened 2021-01-23
[Bhadani Simran] iP

simran-bhadani3:


src/main/java/Duke.java
```diff
@@ -1,10 +1,63 @@
+import java.util.List;
+import java.util.Scanner;
+
 public class Duke {
+
+    /** List of tasks added by the user */
+    private static final List<Task> tasks = Storage.getData();
```
marcusleeeugene, line 8: Perhaps it would be better to rename tasks as TASKS since it is a final variable.

src/main/java/Duke.java
```diff
@@ -1,10 +1,63 @@
...
+                Ui.printInvalidDateFormatMessage();
+            }
+        } else if (Parser.parseCommand(command).equals("delete")) {
+            TaskList.deleteTask(command, tasks);
+        } else if (Parser.parseCommand(command).equals("find")) {
+            TaskList.findTask(command, tasks);
+        } else {
+            throw new InvalidCommandException();
+        }
+    }
```
marcusleeeugene: There is quite some content to go through when looking at these if-else statements. A suggestion would be to use switch cases, and perhaps do a single try-catch block that handles different kinds of exceptions.

src/main/java/Parser.java
```diff
@@ -0,0 +1,83 @@
...
+     * Parses the stored string and finds the completion status.
+     *
+     * @param task String retrieved from Storage.
+     * @return Completion status of the task.
+     */
+    public static String parseCompletionStatus(String task) {
+        String[] split = task.split("\\|");
+        String isDone = split[1].trim();
+        return isDone;
+    }
```
marcusleeeugene, line 53: The variable isDone sounds like a boolean return type instead of a String, perhaps it will be better to rename the variable.

src/main/java/Storage.java
```diff
@@ -0,0 +1,132 @@
...
+    private static void createFile() {
+        try {
+            Files.createFile(fileLocation);
+        } catch (Exception e) {
+            System.out.println(e);
+        }
+    }
+
+    private static boolean checkForFile() {
+        boolean fileExists = Files.exists(fileLocation);
```
marcusleeeugene, line 36: It might sound better as a boolean variable if you rename fileExists variable to isFileExist.

src/main/java/Storage.java
```diff
@@ -0,0 +1,132 @@
...
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.List;
+
+
+public class Storage {
+
+    public static ArrayList<Task> data;
+    private static final Path fileLocation = Paths.get("./data", "duke.txt");
+    private static final Path directoryLocation = Paths.get("./data");
```
marcusleeeugene: Perhaps the variable names could be in upper cases, since they are final variables. It might be difficult for one to differentiate the constant variables.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,184 @@
...
+        System.out.println("Now you have " + count
+                + (count == 1 ? " task" : " tasks")
+                + " in the list.");
+        System.out.println("~~~~~~~~~~~~~~~~~~~~~~~~~~~\n");
+        try {
+            Storage.saveDataToFile(tasks);
+        } catch (UnableToSaveDataException e) {
+            Ui.printUnableToSaveDataMessage();
+        }
+    }
```
marcusleeeugene: Declaration of deleteIndex and count could be placed at the top of the deleteTask method, to easily identify where declared variables are located at.

marcusleeeugene (changes requested):
Overall, the code quality is good, apart from small nits such as the long if-else statements and some variable naming conventions. Great job, keep it up!

src/main/java/Storage.java
```diff
@@ -0,0 +1,132 @@
...
+
+    private static void createFile() {
+        try {
+            Files.createFile(fileLocation);
+        } catch (Exception e) {
+            System.out.println(e);
+        }
+    }
+
+    private static boolean checkForFile() {
```
RuiXiong2211, line 35: Perhaps the method name could be changed to isFileAvailable() to make is sound more like a boolean method.

src/main/java/Storage.java
```diff
@@ -0,0 +1,132 @@
...
+        boolean fileExists = Files.exists(fileLocation);
+        return fileExists;
+    }
+
+    private static boolean checkForDirectory() {
```
RuiXiong2211, line 40: Perhaps the method name could be changed to isDirectoryAvailable() to make is sound more like a boolean method.

src/main/java/Todo.java
```diff
@@ -0,0 +1,21 @@
+public class Todo extends Task {
```
RuiXiong2211, line 4: Perhaps the class name can be changed to ToDo instead.

src/main/java/Duke.java
```diff
@@ -1,10 +1,63 @@
...
+            } catch (InvalidDateTimeFormatException e) {
+                Ui.printInvalidDateFormatMessage();
+            }
+        } else if (Parser.parseCommand(command).equals("delete")) {
+            TaskList.deleteTask(command, tasks);
+        } else if (Parser.parseCommand(command).equals("find")) {
+            TaskList.findTask(command, tasks);
+        } else {
+            throw new InvalidCommandException();
+        }
```
RuiXiong2211: A suggestion would be to use switch statements instead of if else blocks as it would make your code look a alot neater.

src/main/java/Parser.java
```diff
@@ -0,0 +1,83 @@
...
+     * Returns the index of the task to be marked as done.
+     *
+     * @param command Command input by the user.
+     * @return Index of task to be marked as done.
+     */
+    public static int parseDoneIndex(String command) {
+        return Integer.parseInt(command.split(" ")[1]) - 1;
+    }
+
+    private static LocalDateTime DateTimeParser(String input) {
```
RuiXiong2211: Method name should be in  camel case.

RuiXiong2211 (commented):
Good job on adhering to the Java coding standard . I was struggling to find coding violations.

PR #56, opened 2021-01-23
[Marcus Lee Eugene] iP

marcusleeeugene:


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,135 @@
...
+            String date = date(text);
+            commandType = new AddCommand(command, description, date);
+            break;
+        default:
+            throw new DukeException(DukeExceptionType.UNKNOWN_INPUT);
+        }
+        return commandType;
+    }
+
+    private static String command(String text) {
```
yutingzou: Would getCommand or extractCommand be a better naming? 
I've also noticed the same issue in other methods. For example, the 'description', 'date' methods. It might be a bit confusing without a verb in the method name :)

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,109 @@
...
+     * @param command Event or Deadline command
+     * @param description Task name of Event or Deadline
+     */
+    public AddCommand(String command, String description, String date) {
+        super.command = command;
+        super.description = description;
+        super.date = date;
+    }
+
+    private void todoProcess(String taskName, TaskList tasks) {
```
yutingzou, line 44: Since it's an add command, do you think addTodo, addDeadline and addEvent are more informative naming?

I've also noticed the same issue for other Command child classes:) but I like the way you standardised similar method names across all child classes.

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,68 @@
...
+        this.taskList.add(null);
+        this.taskList.addAll(taskList);
+    }
+
+    /**
+     * Size of TaskList, excluding index 0
+     *
+     * @return Size of TaskList
+     */
+    public int size() {
```
yutingzou: Will getSize() be better?

yutingzou (commented):
Your code is very neat, organised and well-commented. 
I've commented some minor method naming issues and please feel free to fix them if you feel appropriate:)

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,109 @@
...
+        storage.save(tasks);
+    }
+
+    /**
+     * Determines if Exit is called by user
+     *
+     * @return false
+     */
+    @Override
+    public boolean isExit() {
```
VisnuRavi: Although this does follow the naming to sound like a boolean, but just a minor issue that isExit as phrase does not seem right to me

src/main/java/duke/Parser.java
VisnuRavi: I do agree with yutingzou on this

VisnuRavi (commented):
Overall I feel your code is great! Apart from a few naming issues, I think your code is good to go!

src/main/java/duke/command/AddCommand.java
marcusleeeugene: Hello, thank you for reviewing my code! I'll work on it :)

src/main/java/duke/Parser.java
marcusleeeugene: Hello, thank you for reviewing my code! Yes I agree, I'll work on it! :)

PR #68, opened 2021-01-24
[Sun Yucheng] iP

Cheng20010201:


src/main/java/lihua/commands/FindCommand.java
```diff
@@ -0,0 +1,33 @@
+package lihua.commands;
+
+/**
+ * Command class representing a command to find tasks by keyword.
+ */
+public class FindCommand extends Command {
+    /** Command word for find command */
```
ZhangAnli: Do you think it would help if you left spaces between consecutive methods? Have noticed this in a few places and because of the javadoc comments, it may make your code seem a bit "cramped".

src/main/java/lihua/tasks/Deadline.java
```diff
@@ -0,0 +1,60 @@
...
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+import org.json.simple.JSONObject;
+
+/**
+ * Represents a lihua.tasks.Deadline item, which is a child of lihua.tasks.Task
+ */
+public class Deadline extends Task {
+    private final LocalDate by;
```
ZhangAnli: Perhaps a more intuitive name for the localdate variable will be better? I feel that just stating "by" could be hard to read for some.

src/test/java/lihua/tasks/DeadlineTest.java
```diff
@@ -0,0 +1,4 @@
+package lihua.tasks;
+
+public class DeadlineTest {
```
ZhangAnli: Would like to clarify the purpose of this file since it is essentially an empty file?

ZhangAnli (commented):
Overall, the code was written well and did not violate much coding violations. Good job!

src/test/java/lihua/tasks/DeadlineTest.java
Cheng20010201: Oops I added the test class but forgot to add content to it. Thanks for pointing it out!

src/main/java/lihua/tasks/Deadline.java
Cheng20010201: like `deadlineDate`? 🤔

src/main/java/lihua/commands/FindCommand.java
Cheng20010201: Sorry correct me if I misunderstood your point🤔. Are you suggesting like I leave spaces of more than one line between consecutive methods to make the code look more 'sparse'?

Eriksen2411: For me, I think the spacing between consecutive methods is good enough. I just think that at the beginning of the class ( between attributes and constructor) you should leave a one line spacing for easier looking

src/main/java/lihua/commands/HelpCommand.java
```diff
@@ -0,0 +1,50 @@
+package lihua.commands;
+
+import lihua.commons.Messages;
+
+/**
+ * Command class representing a command to get help.
+ */
+public class HelpCommand extends Command {
+    /** Command word for help command */
+    public static final String COMMAND_WORD = "help";
```
Eriksen2411: Do we need COMMAND_WORD and MESSAGE_USAGE here at the same time both serving as attributes. I think you should only keep COMMAND_WORD as attribute and the other could be converted to method return.

src/main/java/lihua/tasks/Deadline.java
Eriksen2411: Hmmm i also agree with ZhangAnli this point. and deadlineDate seems good enough.

src/test/java/lihua/tasks/DeadlineTest.java
Eriksen2411: Also agree :)) Just for me to add 1 comment. TY

src/main/java/lihua/tasks/Deadline.java
branzuelajohn: I agree with Anli and Eriksen regarding this

src/main/java/lihua/tasks/Task.java
```diff
@@ -0,0 +1,64 @@
...
+     * Initialize a new undone task
+     * @param name Name of the task
+     */
+    public Task(String name) {
+        this.name = name;
+        isDone = false;
+    }
+
+    /**
+     * Set the status of the task; e.g. done or not done
```
branzuelajohn: "Sets" instead of set?

src/test/java/lihua/tasks/EventTest.java
```diff
@@ -0,0 +1,4 @@
+package lihua.tasks;
+
+public class EventTest {
```
branzuelajohn: Are you supposed to add content to this test class file, because this class does not serve a purpose now as it is empty

src/main/java/lihua/main/Main.java
```diff
@@ -0,0 +1,33 @@
...
+
+import javafx.application.Application;
+import javafx.fxml.FXMLLoader;
+import javafx.scene.Scene;
+import javafx.scene.layout.AnchorPane;
+import javafx.stage.Stage;
+import lihua.ui.MainWindow;
+
+/**
+ * A GUI for Duke using FXML.
```
branzuelajohn: should this be "A GUI for Lihua..." to be consistent with the name you decided to name your bot

PR #73, opened 2021-01-24
[Arihant Jain] iP

arihantjain97:


src/main/java/BaseException.java
```diff
@@ -0,0 +1,10 @@
+public abstract class BaseException extends Exception{
```
marcusleeeugene: The class itself looks fine, however I would suggest leaving an extra spacing after each '{' curly bracket openings.

src/main/java/Duke.java
```diff
@@ -1,10 +1,139 @@
...
+        //overall stringList
+        String[] strArr = new String[100];
+        //overall Task count
+        int count = 0;
+        //overall TaskList
+        ArrayList<Task> TaskList = new ArrayList<>();
+
+        while (sc.hasNextLine()) {
+            try {
+
```
marcusleeeugene: It might look better without an additional newline after each try-catch / if-else statement.

src/main/java/Duke.java
```diff
@@ -1,10 +1,139 @@
...
+                        //task description
+                        String tdStr = str.substring(9, str.indexOf('/') - 1);
+                        strArr[count] = tdStr;
+                        count++;
+                        //task by
+                        String tbyStr = str.substring(str.indexOf('/') + 4, str.length());
+                        Deadlines dlTask = new Deadlines(tdStr, tbyStr);
+                        TaskList.add(dlTask);
+
+
```
marcusleeeugene: The new line format used should perhaps be more consistent in the entire codebase.

src/main/java/DukeException.java
```diff
@@ -0,0 +1,14 @@
+public class DukeException {
+    //error for todo
+    public static class TodoEmpty extends BaseException{
```
marcusleeeugene: The class itself looks fine, however I would suggest leaving an extra spacing after each '{' curly bracket openings.

src/main/java/Task.java
```diff
@@ -0,0 +1,24 @@
...
+        this.description = description;
+        this.isDone = false;
+        count++;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); //return tick or X symbols
+    }
+
+    public void markAsDone(){
```
marcusleeeugene: The class itself looks fine, however I would suggest leaving an extra spacing after each '{' curly bracket openings.

src/main/java/Duke.java
```diff
@@ -1,10 +1,139 @@
...
+                        System.out.println("Got it. I've added this task:");
+                        System.out.println(TaskList.get(count - 1));
+                        System.out.println("Now you have " + count + " tasks in the list.");
+
+
+                    } else if (str.contains("event") && str.contains("/at")) {
+                        //task description
+                        String tdStr = str.substring(6, str.indexOf('/') - 1);
```
marcusleeeugene: Instead of using 'tdStr' to represent task description, maybe you can rename it as 'taskDescription' as it makes the variable more easy to understand.

src/main/java/Duke.java
```diff
@@ -1,10 +1,139 @@
...
+                        strArr[count] = tdStr;
+                        count++;
+                        //task at
+                        String tatStr = str.substring(str.indexOf('/') + 4, str.length());
```
marcusleeeugene: Instead of using 'tatStr' to represent task at, maybe you can rename it as 'taskAt' as it makes the variable more easy to understand.

src/main/java/Task.java
```diff
@@ -0,0 +1,24 @@
+public class Task {
+    public String description;
+    public boolean isDone;
```
marcusleeeugene: I like how the boolean 'isDone' really does sound like a boolean variable. It is easy to differentiate it from other type of variables.

marcusleeeugene (changes requested):
Overall, I think the coding standards are great, apart from a few nits, such as consistency in new lines, line spacing after curly brackets and variable namings. Good job!

PR #77, opened 2021-01-24
[Chew Wei Hao] iP

chewwh09:


src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,56 @@
...
+ * The AddCommand class denotes an add command to the Duke chat bot.
+ */
+public class AddCommand extends Command {
+    private final String type;
+
+    /**
+     * Constructs an AddCommand.
+     * @param type              The type of task to be added into the list.
+     * @param taskDescription   The description of the task.
+     */
```
onnwards, line 21: Should there be an empty line between description and parameter section for javadocs? I noticed this for your other javadocs as well.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,62 @@
...
+import duke.logging.TaskList;
+import duke.logging.Ui;
+
+/**
+ * A personal chat bot that helps a person keep track of various things.
+ */
+public class Duke {
+    private final Storage storage;
+    private TaskList tasks = new TaskList(new ArrayList<>());
+    private final Ui ui;
```
onnwards, line 18: ### Should `storage` and `ui` be in uppercase?

Honestly, I wouldn't consider this to be violations, but I'll leave it to you to judge: 
    - Your variables are final whenever possible, which is good, 
    - However, that means it probably should be written in uppercase according to the coding convention. 
    - My personal belief is that only static final variables should be written in uppercase, but the coding convention isn't clear on that. 
    - However, an example they provided was `final int COLOR_RED = 1;` which leans towards the former,

### I would leave them as non-final to avoid meeting this problem.

onnwards (commented):
### LGTM, with a few minor coding standard violations. I really liked how you indented all the javadoc params and returns to start at the same level.

I know I'm only supposed to be commenting on coding standard violations, but some other things I noticed:
Inconsistent use of `@Override` on `isExit()`
Will `"data/duke.txt"` have cross platform issues? (Windows uses `\`)

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,49 @@
...
+     * Executing the command
+     * @param taskList The list of recorded tasks.
+     * @param ui       The user interface.
+     * @param storage  The list of recorded user inputs data.
+     * @return         The message replied by Duke chat bot.
+     */
+    @Override
+    public String execute(TaskList taskList, Ui ui, Storage storage) {
+        try {
+            int index = Integer.parseInt(taskDescription.substring(0, 1)) - 1;
```
icytornado, line 36: maybe can not more explanatary variable here,
which can better explain the purpose of variable

src/main/java/duke/logging/Parser.java
```diff
@@ -0,0 +1,50 @@
...
+        if (fullCommand.contains(" ")) {
+            int commandStrLength = fullCommand.indexOf(" ");
+            type = fullCommand.substring(0, commandStrLength);
+            taskDescription = fullCommand.substring(commandStrLength + 1, fullCommand.length());
+        } else {
+            type = fullCommand;
+            taskDescription = "";
+        }
+
+        if (type.equals("list")) {
```
icytornado: maybe can replace nested if-else with cases, and a default?

src/main/java/duke/logging/TaskList.java
```diff
@@ -0,0 +1,155 @@
...
+     */
+    public TaskList(ArrayList<Task> tasks) {
+        this.tasks = tasks;
+    }
+
+    /**
+     * Print out all the tasks in the list.
+     */
+    public String list() {
+        String message = "     Here are the tasks in your list:";
```
icytornado, line 34: maybe can leave a blankline after this to group similar codes together for better readability?

src/main/java/duke/logging/TaskList.java
```diff
@@ -0,0 +1,155 @@
...
+            String taskName = taskDescription.substring(0, index);
+            String dateTimeString = taskDescription.substring(index + 4).strip()
+                    .replace("/", "-");
+            LocalDate dateTime = LocalDate.parse(dateTimeString);
+
+            if (type.equals("deadline")) {
+                task = new Deadline(false, taskName, dateTime);
+            } else {
+                task = new Event(false, taskName, dateTime);
+            }
```
icytornado, line 76: would it be better to extract out this nested if-else statement,
as a method, so that the reader do not need to trace the code too much?

src/main/java/duke/logging/TaskList.java
```diff
@@ -0,0 +1,155 @@
...
+     * Add a task into the list of tasks.
+     * @param type                         The type of task.
+     * @param taskDescription              The description of the task.
+     * @param storage                      The storage information containing the list of tasks information.
+     * @return                             The updated list of tasks.
+     * @throws InvalidDescriptionException If description format is wrong.
+     * @throws DateTimeParseException      If datetime format is wrong.
+     * @throws IOException                 If inputting information into the storage post an error.
+     */
+    public ArrayList<Task> addTask(String type, String taskDescription, Storage storage)
```
icytornado, line 51: would it be better if a noun is used,
instead of verb. For example, can be taskAdder?

icytornado (commented):
Overall, nicely written code,
with good quality, not big violation of the code quality guidelines.
Keep it up!

PR #79, opened 2021-01-24
[Yuen Sheung Man] iP

shermanyuen:


src/main/java/Duke.java
```diff
@@ -1,10 +1,148 @@
...
+            userInput = scan.nextLine();
+            parsedUserInput = new SplitString(userInput, " ");
+            userCommand = parsedUserInput.getFirstString().toLowerCase();
+            userDescription = parsedUserInput.getSecondString();
+
+            if (String.valueOf(userCommand).equals("bye") == true) {
+                break;
+            }
+            switch (userCommand) {
+                case "list":
```
douglaswja: The 'case' statements inside a 'switch' statement should be aligned to the same indentation as the 'switch' statement.

If IntelliJ is your IDE, you can follow the instructions at the following link for how to get IntelliJ to follow this coding convention for you.
https://stackoverflow.com/questions/40888798/config-intellij-switch-case-style#:~:text=Go%20to%20Settings%20%3E%20Editor%20%3E%20Code,Indent%20'case'%20branches%22.


For example:
```
switch (userCommand) {
case "list":
  // Code
  break;
case "done":
  // Code
  break;
...

```

src/main/java/Task.java
```diff
@@ -0,0 +1,31 @@
...
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "Y" : " ");
```
douglaswja, line 11: This ternary statement could be rewritten to match the CS2103 Java Coding Standard.

From the CS2103 Java Coding Standard (https://se-education.org/guides/conventions/java/intermediate.html)
Here are two acceptable ways to format ternary expressions:
```

alpha = (aLongBooleanExpression) ? beta : gamma;
alpha = (aLongBooleanExpression)
        ? beta
        : gamma;

```

For example, perhaps the following might follow the convention more closely:
``` return (isDone) ? "Y" : " ";```

**Comment by jingxueguo** [comment]
Hello! I think it is better if you use present tense instead of past tense in your commit messages, e.g. Add Level-2 requirements rather than Added Level-2 requirements

File: src/main/java/Duke.java [hunk 3]
```diff
@@ -1,10 +1,148 @@
...
+        if (String.valueOf(userDescription).equals("") == true) {
+            printMessage("Oops, the description of a '" + userCommand + "' cannot be empty.");
+            return;
+        }
+        Task t;
+        String taskDescription;
+        SplitString parsedUserDescription;
+        
+        switch (userCommand) {
+            case "todo":
```
douglaswja: As in the previous comment, the case statements should follow the same indentation as the switch.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
+public class Deadline extends Task {
+    protected String by;
```
jingxueguo: do you think maybe a noun will be a better variable name here? maybe something like "dueDate"?

douglaswja (commented):
Just a few nits here and there.

src/main/java/Duke.java
```diff
@@ -1,10 +1,148 @@
+import java.util.Scanner;
+
 public class Duke {
-    public static void main(String[] args) {
+
+    public static Task[] tasks = new Task[100];
```
yanlingkuek: maybe an arraylist can be used instead so that there will be no issues of exceeding the array size

src/main/java/Duke.java
```diff
@@ -1,10 +1,148 @@
...
+        String welcomeMessage = "Hello from \n" + logo + "\n"
+                + "What can I do for you?";
+        printMessage(welcomeMessage);
+    }
+
+    public static void printGoodbyeMessage() {
+        printMessage("Bye. Hope to see you again soon!");
+    }
+
+    public static void listTasks() {
```
yanlingkuek: it may be a bit neater to move methods like `listTasks, addTask, deleteTask` etc. into a separate class called `TaskList` maybe

shermanyuen: Thank you for your comments! As it turns out, this suggestion is also provided in `A-MoreOOP`, but I haven't reached there yet. I was still at Level-6. I will follow the task requirements now.

src/main/java/Duke.java
shermanyuen: The task description in `Level-2` says that we can assume at most 100 tasks. Anyway, how to implement an arraylist?

src/main/java/Task.java
shermanyuen: The ternary statement in the code matches the first acceptable way (that the whole ternary expression is in one line). I think it is redundant to include brackets for `isDone`, i.e. `return ((isDone) ? "Y" : " ");` is redundant since the boolean expression is a short one.

PR #83, opened 2021-01-24
[Song Soon Wee] iP

ssoonwee:


src/main/java/soonwee/duke/Deadline.java
```diff
@@ -0,0 +1,30 @@
...
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents a Deadline instance. A Deadline instance will store the task
+ * name and its due completion time.
+ */
+public class Deadline extends Task {
+
+    protected LocalDateTime by;
```
mesyeux: Perhaps you could consider using a more specific name such as `dateBy` so it is clearer to readers?

src/main/java/soonwee/duke/Deadline.java
```diff
@@ -0,0 +1,30 @@
...
+     */
+    public Deadline(String description, LocalDateTime by) {
+        super(description);
+        this.by = by;
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + "(by: "
+                + by.format(DateTimeFormatter.ofPattern("MMM dd yyyy HH:mm")) + ")";
```
mesyeux: Great indentation to improve readability!

src/main/java/soonwee/duke/Duke.java
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * Represents a Duke instance. A Duke instance will prompt user which bot they
+ * want to use (1. Echo Bot, 2. Tasks List Bot).
+ */
+public class Duke {
+
+    public static final int TODO = 1;
+    public static final int DEADLINE = 2;
+    public static final int EVENT = 3;
+    public static final int FIND = 4;
```
mesyeux: Great practice of capitalising the constant variables. Just curious but is there a reason why you chose to use constant variables over enumerations?

src/main/java/soonwee/duke/Duke.java
```diff
@@ -0,0 +1,160 @@
...
+                BufferedReader(new InputStreamReader(System.in));
+        while (true) {
+            cmd = input.readLine();
+            System.out.println(display.printLine());
+            if (cmd.equals("list")) {
+                storage.taskList.displayTasks();
+            } else if (cmd.equals("bye")) {
+                break;
+            } else if (cmd.contains("done")) {
+                int index = Integer.parseInt(cmd.substring(5));
```
mesyeux: Perhaps you could specify what this `index` is referring to and rename it as such. Maybe `indexOfDescription?`

src/main/java/soonwee/duke/Duke.java
```diff
@@ -0,0 +1,160 @@
...
+                        performAddTask(taskList, newToDo);
+                    } else {
+                        LocalDateTime date = checker.dateFormatter(cmd);
+                        if (taskType == DEADLINE) {
+                            Deadline newDeadLine = new Deadline(task, date);
+                            performAddTask(taskList, newDeadLine);
+                        } else {
+                            Event newEvent = new Event(task, date);
+                            performAddTask(taskList, newEvent);
+                        }
```
mesyeux: This looks a little like arrowhead style code. Perhaps you can consider simplifying it so that it's less complicated and indented.

src/main/java/soonwee/duke/Storage.java
```diff
@@ -0,0 +1,128 @@
...
+     * Instantiate Storage with its target directory.
+     */
+    public Storage(String directory) {
+        this.taskList = createNew();
+        this.directory = directory;
+        this.fileObject = createFile(directory);
+    }
+
+    /**
+     * Create new TaskList instance for this instance.
```
mesyeux: Maybe you can consider using **Creates** instead of **Create**

src/main/java/soonwee/duke/Storage.java
```diff
@@ -0,0 +1,128 @@
...
+        } catch (IOException e) {
+            System.out.println("An error has occurred.");
+            e.printStackTrace();
+        }
+        return fileObject;
+    }
+
+    /**
+     * Read file contents.
+     */
```
mesyeux: Perhaps you could leave this JavaDoc comment out, as it seems to be repeating the obvious.

mesyeux (commented):
Overall I thought your code was quite easy to read with informative names and good checks. Some of your code and methods are slightly long and therefore harder to read/understand so perhaps you could look at that.

src/main/java/soonwee/duke/Duke.java
skinnychenpi: Same question

src/main/java/soonwee/duke/Duke.java
```diff
@@ -0,0 +1,160 @@
...
+    /**
+     * Performs the second macro task which can add, delete or mark a task as
+     * done. Macro will end when user types in 'bye'.
+     *
+     * @return updated Task List.
+     * @throws IOException if file or directory is not found.
+     */
+    public static TaskList performSecondTask() throws IOException {
+        String cmd = new String();
+        Storage storage = new Storage("C:\\Users\\songs\\Desktop\\CS2103\\data\\tasks.txt");
```
skinnychenpi: Just curious whether using absolute path will influence future distribution of the program.

src/main/java/soonwee/duke/Duke.java
skinnychenpi: Maybe it's better to check the taskType, for example:
`if (taskType == TODO) { 
.....} else if (taskType == DEADLINE) {
.....} else if (taskType == EVENT) {
.....} else {
throw new DukeException(....);
}
`

src/main/java/soonwee/duke/Storage.java
```diff
@@ -0,0 +1,128 @@
...
+        if (taskType == 'T') {
+            this.taskList.addTask(new ToDo(task));
+        } else {
+            LocalDateTime formatDate = LocalDateTime.parse(time, formatter);
+            if (taskType == 'D') {
+                this.taskList.addTask(new Deadline(task, formatDate));
+            } else {
+                this.taskList.addTask(new Event(task, formatDate));
+            }
+        }
```
skinnychenpi: Perhaps it's better not to use too much else statement as there could be some other types of tasks in the future and so it's harder to maintain the code. Maybe use more else if to judge the type of tasks will be better.

src/main/java/soonwee/duke/Storage.java
```diff
@@ -57,6 +78,11 @@ public void readFile() {
         }
     }
 
+    /**
+     * Process data in file and add them into memory task list.
+     *
+     * @param data input data taken from command.
+     */
```
skinnychenpi: Maybe you can consider to change Process into Processes. Just one tiny suggestion.

src/main/java/soonwee/duke/Duke.java
ssoonwee: Yes I totally agree with this. I have created an enum class to resolve this. Thanks :)

PR #85, opened 2021-01-24
[Li Jiahe] iP

ljhgabe:


src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,44 @@
...
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException {
+        int index = getIndex();
+        if (index < tasks.size()) {
+            Task removingTask = tasks.get(index);
+            tasks.remove(index);
+            storage.updateInFile(tasks);
+            System.out.println(" Following task is removed:");
+            System.out.println("  " + removingTask);
+            System.out.println(" Now you have " + tasks.size() + " tasks.");
```
pyuxiang: You have a `Ui` object passed to the command, which ideally handles message passing to `System.out`.
Can consider offloading these to the `ui` instead.

Same as in the other commands.

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Deletes the specified task from the task list.
+     * @param tasks Task list given.
+     * @param ui User interface class object.
+     * @param storage Storage path that is going to be updated.
+     * @throws DukeException If error occurs during the process.
+     */
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException {
```
pyuxiang: This line here states that `DukeException` is thrown, but only `TaskIndexOutOfBoundException` is actually thrown.
This will require the stack to explicitly handle the general exception instead of a more specific one.
Consider changing it to the `throws TaskIndexOutOfBoundException`?

Same goes to the other commands.

src/main/java/duke/command/EventCommand.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    protected Event getTask() throws DescriptionMissingException, InvalidDateTimeException {
+        String nameDate = fullCommand.substring(5).strip();
+        if (nameDate.equals("")) {
+            throw new DescriptionMissingException("Argument missing!");
+        }
+
+        String[] nameAndDate = nameDate.split("/at");
+        if (nameAndDate.length < 2) {
+            throw new DescriptionMissingException("Argument missing!");
```
pyuxiang: I couldn't immediately tell what this error is, perhaps more so from the user perspective.
A more descriptive error message might help, e.g. `"Argument missing! Event needs to be supplied a date."`.

src/main/java/duke/command/ExitCommand.java
```diff
@@ -0,0 +1,30 @@
...
+     * @param ui User interface class object.
+     * @param storage Storage path that is going to be updated.
+     */
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        ui.sayGoodBye();
+    }
+
+    /**
+     * Returns true.
```
pyuxiang: Should describe what the function is intended to achieve, instead of stating how.
Perhaps, "Passes instruction to main routine to exit the program"

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,107 @@
...
+        } else if (potentialCommand.equalsIgnoreCase("FIND")) {
+            command = new FindCommand(input);
+        } else {
+            throw new CommandNotFoundException("What do you mean? I do not know this command.");
+        }
+        return command;
+    }
+
+    /**
+     * Returns a LocalDateTime from a string containing date and time.
```
pyuxiang: It's not immediately clear how the datetime is parsed.
You might want to add in a description of how this parsing is done as well, as an
additional paragraph under the method summary.

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,72 @@
...
+     * @param tasks The TaskList to be printed.
+     */
+    public void listTasks(TaskList tasks) {
+        System.out.println(" Here are the tasks: ");
+        for (int i = 0; i < tasks.size(); i++) {
+            Task tempTask = tasks.get(i);
+            System.out.println(" " + (i + 1) + "." + tempTask);
+        }
+    }
+
```
pyuxiang, line 104: You probably inadvertently missed out commenting this public class.

pyuxiang (commented):
Is adding of the various JUnit libraries to the repository intended?
If it's to tie in to a particular version, these are already specified in your Gradle configuration.

Otherwise codebase is very well-designed 👍

src/main/java/duke/command/DeleteCommand.java
ljhgabe: > This line here states that `DukeException` is thrown, but only `TaskIndexOutOfBoundException` is actually thrown.
> This will require the stack to explicitly handle the general exception instead of a more specific one.
> Consider changing it to the `throws TaskIndexOutOfBoundException`?
> 
> Same goes to the other commands.

Thanks for pointing it out! That is indeed a very good suggestion.

src/main/java/duke/command/DeleteCommand.java
ljhgabe: I also realised this and I will make some modification later. Thanks for pointing it out!

src/main/java/duke/command/EventCommand.java
ljhgabe: Good point! I was too lazy to add this descriptive message:(. I will make it more descriptive later.

src/main/java/duke/command/ExitCommand.java
ljhgabe: Thanks!

src/main/java/duke/parser/Parser.java
ljhgabe: Will add and many thanks!

src/main/java/duke/ui/Ui.java
ljhgabe: Yeah exactly... Thanks for pointing it out!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * An empty constructor for the Duke bot.
+     */
+    public Duke() {
+    }
+    /**
+     * Constructs a Duke bot.
+     * @param filePath The file path to store the tasks.
+     */
```
riyadh-h, line 20: A minor detail, but shouldn't there be an empty line between the method's description and parameters in the JavaDoc comment?

I noticed this in other parts of the code as well.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * Main method to lunch the bot.
+     * @param args Arguments of the main method.
+     */
+    public static void main(String[] args) {
+        new Duke(System.getProperty("user.dir") + "/duke.txt").run();
+    }
+
+    /**
+     * Launch the Duke bot by reading in the user input and gives responses.
```
riyadh-h, line 39: Shouldn't it be `Launches` instead of `Launch`?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,76 @@
...
+            } catch (DukeException e) {
+                ui.showErrorMessage(e.getMessage());
+            }
+        }
+    }
+
+    /**
+     * You should have your own function to generate a response to user input.
+     * Replace this stub with your completed method.
+     */
```
riyadh-h, line 61: Shouldn't this JavaDoc comment be updated?

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,26 @@
...
+
+import duke.exception.DukeException;
+import duke.storage.Storage;
+import duke.task.TaskList;
+import duke.ui.Ui;
+
+/**
+ * An abstract class that represents a Command.
+ */
+public abstract class Command {
```
riyadh-h, line 11: I like how you have created an abstract `Command` class 👍.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,33 @@
...
+ * An abstract class that represents "adding task" command.
+ */
+public abstract class AddCommand extends Command {
+    /**
+     * Adds the task into a given task list while updating the file.
+     * @param tasks Task list.
+     * @param task Task to be added.
+     * @param storage Storage path that is going to be updated.
+     */
+    public void addThisTask(TaskList tasks, Task task, Ui ui, Storage storage) {
```
riyadh-h, line 31: Wouldn't it be more concise for the method name if it was `addTask` instead?

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,57 @@
...
+            }
+        }
+        ui.listMatchingTasks(matchingTasks);
+        return ui.listMatchingTasksResponse(matchingTasks);
+    }
+
+    /**
+     *
+     * @return False.
+     */
```
riyadh-h: Shouldn't this JavaDoc comment include a description?

src/test/java/duke/task/DeadlineTest.java
```diff
@@ -0,0 +1,22 @@
...
+        Deadline deadline = new Deadline("Homework 5", deadlineDateTime);
+        assertEquals("[D][✘] Homework 5 (by: Jan 27 2021 23:59)", deadline.toString());
+    }
+    @Test
+    public void testToFileString() {
+        LocalDateTime deadlineDateTime = LocalDateTime.of(2021, 01, 27, 23, 59);
+        Deadline deadline = new Deadline("Homework 5", deadlineDateTime);
+        assertEquals("D | 0 | Homework 5 | 2021-01-27 2359", deadline.toFileString());
+    }
+}
```
riyadh-h: Shouldn't the names of the test methods be more descriptive? (e.g., by using the three part format)

riyadh-h (commented):
Overall, the code is well organised and easy to read. There are a few minor details, however, that I think they should be reviewed. 👍

PR #86, opened 2021-01-24
[Tran Nguyen Manh Dan] iP

dannn-tran:


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,59 @@
...
+     * Creates a Duke instance with filePath to where data is stored.
+     *
+     * @param filePath
+     */
+    public Duke(String filePath) throws IOException {
+        storage = new Storage(filePath);
+        ui = new Ui();
+        try {
+            taskList = storage.load();
+        } catch (DukeException e) {
```
jamesleeht, line 29: According to coding standard, it's not good to have single letter variables. Maybe this can be called "exception" instead?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,89 @@
...
+     */
+    public static final Command parse(String userInput) throws DukeException {
+        if (userInput.equals("list")) {
+            return new ListCommand();
+        } else if (userInput.equals("bye")) {
+            return new ExitCommand();
+        } else {
+            String[] splitInput = userInput.split(" ", 2);
+            switch (splitInput[0]) {
+
```
jamesleeht: According to coding standard, there shouldn't be an extra line of space here.

jamesleeht (commented):
Coding standard closely followed except for one or two lines, nice job.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,77 @@
...
+
+public class Storage {
+    private File file;
+
+    /**
+     * Creates necessary directories to the file path if not present
+     *
+     * @param filePath
+     * @return TaskList
+     * @throws DukeException
```
goatygoatygoat, line 21: Perhaps you could describe the circumstances which would cause a DukeException to be thrown?

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,35 @@
+package duke.command;
+
+import duke.DukeException;
+import duke.Storage;
+import duke.task.Task;
+import duke.task.TaskList;
+
+public class DeleteCommand extends Command {
+    private int idx;
```
goatygoatygoat, line 9: Perhaps you can use the full word "index" as the variable name instead?

goatygoatygoat (commented):
Looks good 👍 
Sorry for reviewing so late.

PR #88, opened 2021-01-24
[V R Soorya] iP

VRSoorya:
Pull Request from forked repo's master to upstream repo's master


src/main/java/project/OlafApp.java
```diff
@@ -0,0 +1,174 @@
...
+     * Executes each session of the application.
+     */
+    public void run() {
+        // todo: break this down into more methods...?
+        BufferedReader bf = new BufferedReader(new InputStreamReader(System.in));
+
+        System.out.println(PrintText.WELCOME_MESSAGE);
+
+        while (isActive) {
+            String command = null;
```
samuelfangjw: Perhaps this could be an enumeration instead?

src/main/java/project/OlafApp.java
```diff
@@ -0,0 +1,174 @@
...
+            if (command.equalsIgnoreCase("bye")) {
+                this.stop();
+            } else if (command.equalsIgnoreCase("list")) {
+                // todo: use try catch here
+                if (taskList.hasTasks()) {
+                    ui.showList(taskList);
+                } else {
+                    System.out.println(PrintText.EMPTY_TASKLIST_ERROR);
+                }
+            } else if (command.toLowerCase().startsWith("find")) {
```
samuelfangjw: Many of your cases start with `command.toLowerCase()`, should this could be abstracted out as a variable instead?

src/main/java/project/OlafApp.java
```diff
@@ -0,0 +1,174 @@
...
+        this.taskList = tasks;
+        this.ui = ui;
+        this.storage = storage;
+        this.isActive = true;
+    }
+
+    /**
+     * Executes each session of the application.
+     */
+    public void run() {
```
samuelfangjw: This method seems a little long. Would it be better to abstract out some of the logic into other methods instead?

src/main/java/project/OlafApp.java
```diff
@@ -0,0 +1,174 @@
...
+                    ui.showNewTaskAddedSuccess(taskList.getTotalNumberOfTasks(),
+                            newEvent, taskList.getTotalNumberOfTasksUndone());
+                } catch (Exception e) {
+                    // catches both ParseException and IndexOutOfBounds exception
+                    ui.showFormatError(PrintText.EVENT_FORMAT);
+                }
+            } else if (command.equalsIgnoreCase("help")) {
+                ui.showFormatResponse(PrintText.HELP);
+            } else {
+                ui.showFormatResponse("  Hmm sorry I don't understand :(\n"
```
samuelfangjw: Perhaps this response could be added to your PrintText class?

samuelfangjw (commented):
Code was easy to read and looks great! I love that you put in the effort and time to try new things and I learnt a few tricks from reading your code! Overall the program seems well structured and I had a enjoyable time reviewing it!

whatthelump:
Code closely adhered to the given standards; I am impressed by the attention to detail and found the code to be very pleasant to read.

**Naming**

- Names representing packages (project.task) are all in lower case. 
- Class names are all nouns and written in PascalCase.
- Variable names are all in camelCase. Methods are all verbs and written in camelCase. Boolean variables sound like Booleans (isDone, hasTasks, taskIsDone). Plural nouns are used for collections (tasks, which is a collection of Tasks).

**Layout**

- Code follows indentation guidelines of 4 spaces for basic indentations and 8 spaces (from parent line) for wrapped lines. 
- Line length did not exceed 120 characters. 
- There are consistent breaks after a comma and before an operator. 
- Method/constructor names stay attached to the following open parenthesis “(“. 
- Egyptian style brackets were applied consistently. 
- Appropriate white space usage within statement: operators surrounded by a space character, and Java reserved words such as if are followed by a white space.

**Statements**

- Every class is part of the ‘project’ package. 
- There is consistent ordering of import statements (e.g. java.time.format.DateTimeFormatter always follows java.time.LocalDateTime). 
- Imported classes are listed explicitly, there is no usage of wildcard imports. 
- Most variables are initialized when declared.
- Class variables are all either private or protected.

**Comments**

- You provided descriptive header comments for all classes, and even for methods like getters where it is not strictly necessary.
- You followed Javadoc form to a tee as far as I can see; all method summaries began with verbs with an ‘s’ at the end, parameters were all listed and there was even proper punctuation after each parameter description, which shows meticulousness as this seems easy to overlook.

whatthelump:
https://github.com/VRSoorya/ip/blob/dddf901cc98403f3129d06174e6ed93ef89abc74/src/main/java/project/common/PrintText.java#L6

For enum name PrintText, perhaps you could change it to something like ‘PrintedText’ to make it sound more like a noun?

whatthelump:
https://github.com/VRSoorya/ip/blob/dddf901cc98403f3129d06174e6ed93ef89abc74/src/main/java/project/task/TaskList.java#L12-L13

I think this is trivial but perhaps you could initialize the variable tasks under class TaskList to new ArrayList<Task>()? Then you might be able to do away with the empty constructor.

src/main/java/project/task/TaskList.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * Represents a list of {@code Task}s in the application.
+ * Encapsulates storage and operations on {@code Task}s.
+ */
+public class TaskList {
+    private List<Task> tasks;
```
whatthelump, line 15: I think this is trivial but perhaps you could initialize the variable tasks under class TaskList to new ArrayList()? Then you might be able to do away with the empty constructor.

src/main/java/project/common/PrintText.java
```diff
@@ -0,0 +1,90 @@
+package project.common;
+
+/**
+ * All the {@code String} constants commonly used in the application.
+ */
+public enum PrintText {
```
whatthelump: For enum name PrintText, perhaps you could change it to something like ‘PrintedText’ to make it sound more like a noun?

VRSoorya: Right I hadn't noticed that, thanks!

src/main/java/project/task/TaskList.java
VRSoorya: That's true, thanks for the suggestion! I'll make the necessary changes 👍

src/main/java/project/OlafApp.java
VRSoorya: Yup, I agree!

src/main/java/project/OlafApp.java
VRSoorya: I'm thinking about it too! I've got some ideas but will have to refactor things and see how it goes :)

src/main/java/project/OlafApp.java
VRSoorya: Oh yup that's true!

src/main/java/project/OlafApp.java
VRSoorya: Do you mean an enumeration for the commands so I can check against those in each 'if' condition here? I didn't think of it earlier but yes that's a good suggestion! will look into it 👍

PR #93, opened 2021-01-24
[Lan Yu Xuan] iP

Jellybeano:


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,749 @@
...
+    private ScrollPane scrollPane;
+    private VBox dialogContainer;
+    private TextField userInput;
+    private Button sendButton;
+    private Scene scene;
+    private Image user = new Image(this.getClass().getResourceAsStream("/User.jpg"));
+    private Image duke = new Image(this.getClass().getResourceAsStream("/Duwuke.jpg"));
+
+    static class FastIO extends PrintWriter
+    {
```
chwoozy: Perhaps you may want to change this to Egyptian style braces (CS2103T coding style)?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,749 @@
...
+            }
+            return str;
+        }
+    }
+
+    public static class Task {
+        /** The task that needs to be done, as a String */
+        protected String todo;
+        /** Whether or not the task is complete */
+        protected boolean done;
```
chwoozy: Maybe you want to change the boolean variable name to sound like a boolean instead? How about "isTaskDone", how does that sound? 😄

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,749 @@
...
+        /**
+         * Loads a TaskList from storage, if it exists.
+         * Else creates a new storage file.
+         *
+         * @return the TaskList stored on the hard disk.
+         */
+        public TaskList load() {
+            File taskData = new File(filePath);
+
+            TaskList leest = new TaskList();
```
chwoozy: Haha, I get the joke! But maybe it would be better to name it to a more readible and intuitive variable name, like "taskList". In case some developers do not appreciate the humor. 😝

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,749 @@
...
+         *
+         * @param leest The current TaskList.
+         * @param input Input that is being read in.
+         */
+        public String parse(TaskList leest, String input) {
+
+            String[] split = input.split("\\s+");
+
+            switch (split[0]) {
+            case "list":
```
chwoozy: I like your use of switch case statements! However, do note that we may need to add the "//Fallthrough" under each case where a break  is supposed to be placed, but unreachable!

chwoozy (commented):
Code generally clean apart from a few nits! LGTM!

src/main/java/duke/Duke.java
```diff
@@ -384,85 +406,78 @@ public void remove(Task t) {
...
         /**
-         * Parses a given input and performs actions as necessary based on the input.
+         * Parses a given input and returns Strings as necessary based on the input.
          * The main driver of the Duke class.
          *
          * @param leest The current TaskList.
          * @param input Input that is being read in.
          */
-        public void parse(TaskList leest, String input) {
+        public String parse(TaskList leest, String input) {
```
shaelynl: Haha cute name, but maybe you could consider changing it to something more understandable like taskList?

src/main/java/duke/Duke.java
```diff
@@ -594,30 +599,29 @@ public void showList(TaskList tl) {
...
+        public String showSearchList(TaskList tl) {
             if (tl.isEmpty()) {
-                System.out.println("There are no matching tasks uwu. "
+                return ("There are no matching tasks uwu. "
                         + "(just like how whoever wrote this has no friends.)");
             } else {
-                System.out.println("Here are the matching tasks in your list uwu: ");
+                String toReturn = ("Here are the matching tasks in your list uwu: ");
                 int counter = 1;
                 for (Task t : tl.getList()) {
```
shaelynl: I think under the Java coding standards the convention for a for loop is for (initialization; condition; update) (although I also use enhanced loops), you could consider changing it?

shaelynl (commented):
Your code is generally clean and easy to understand. Great job! 😃

PR #94, opened 2021-01-24
[Wang Tao] iP

wangtao0717:


src/main/java/duck/Duke.java
```diff
@@ -0,0 +1,46 @@
+package duck;
+
+import duck.operation.Command;
+import duck.operation.Parser;
+import duck.operation.Storage;
+import duck.operation.Ui;
+import duck.task.TaskList;
+
+import java.io.*;
```
kouyk: Preferably java.* should be above any third-party packages

src/main/java/duck/Duke.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class Duke {
+    private Storage storage;
+    private TaskList tasks;
+    private Ui ui;
+
+    /**
+     * initialize Duck object
+     * @param filePath the path of the file containing data of tasks
```
kouyk: Please leave a line break between the description and parameter section.

src/main/java/duck/Duke.java
```diff
@@ -0,0 +1,46 @@
...
+        do {
+            fullCommand = ui.readCommand();
+            Command c = Parser.parse(fullCommand);
+            c.execute(tasks, ui, storage);
+
+        } while (!fullCommand.equals("bye"));
+    }
+
+    public static void main(String[] args) throws IOException {
+        String filePathOfData = ".\\data\\duke.txt";
```
kouyk: Do you think it will be better to name the String as `dataFilePath`?

src/main/java/duck/operation/Command.java
```diff
@@ -0,0 +1,148 @@
...
+                ui.showErrorReply("error_done_no_meaning");
+            } catch (IndexOutOfBoundsException e) {
+                ui.showErrorReply("error_done_non_existed_task");
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+            break;
+        case "delete":
+            try {
+                tasks.getTask(Integer.parseInt(description) - 1);
```
kouyk: Will it be better if `Integer.parseInt(description) - 1` is extracted as a variable first since it will be used multiple times?

src/main/java/duck/operation/Command.java
```diff
@@ -0,0 +1,148 @@
...
+            descriptionSplit = description.split("/by");
+            try {
+                tasks.addTask(new Deadline(descriptionSplit[0].trim(), descriptionSplit[1].trim()));
+                storage.updateFile(tasks);
+                ui.showCommandReply(command, tasks);
+            } catch (ArrayIndexOutOfBoundsException e) {
+                ui.showErrorReply("error_deadline_by");
+            } catch (DateTimeParseException e) {
+                ui.showErrorReply("error_deadline_by");
+            }
```
kouyk: This is very similar to the `event` case, will it be better to extract this as a function?

src/main/java/duck/operation/Command.java
```diff
@@ -0,0 +1,148 @@
...
+            break;
+        case "find":
+            try {
+                String[] result = tasks.findTask(description);
+                ui.showFindReply(result);
+            } catch (NullPointerException e) {
+                ui.showErrorReply("find_empty");
+            }
+            break;
+        default:
```
kouyk: I like how you are using the `default` branch to catch errors.

src/main/java/duck/operation/Storage.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * initialize Storage object, location file
+     * if it does not exist, generate a new one
+     *
+     * @param filePath the path of file
+     * @throws IOException
+     */
+    public Storage(String filePath) throws IOException {
+        this.fileOfData = new File(filePath);
```
kouyk, line 26: The use of `this` is to be avoided according to the style guide.

src/main/java/duck/operation/Storage.java
```diff
@@ -0,0 +1,62 @@
+package duck.operation;
+
+import duck.task.TaskList;
+
+import java.io.*;
+
+public class Storage {
+    private File fileOfData;
```
kouyk, line 17: Naming the variable as \*Of\* is rather awkward, do you think `dataFile` is simpler and straight to the point?

src/main/java/duck/operation/Storage.java
```diff
@@ -0,0 +1,62 @@
...
+     *
+     * @return a list of data
+     * @throws IOException
+     */
+    public String[] load() throws IOException {
+        FileInputStream inStream = new FileInputStream(fileOfData);
+        InputStreamReader reader = new InputStreamReader(inStream);
+        BufferedReader br = new BufferedReader(reader);
+        String[] data = new String[1000];
+        int lines = 0;
```
kouyk, line 50: I think this should be `lineNumber` right? It is pretty misleading on the first look.

src/main/java/duck/operation/Ui.java
```diff
@@ -0,0 +1,156 @@
+package duck.operation;
+
+import duck.task.TaskList;
+
+import java.util.Scanner;
+
+public class Ui {
+    private static String INDENTATION = "    ";
+    private static String HORIZON = "------------------------------------------------------";
```
kouyk: Do you mean `HORIZONTAL_LINE`?

src/main/java/duck/operation/Ui.java
```diff
@@ -0,0 +1,156 @@
...
+            System.out.println(INDENTATION + "OOPS!!! The deadline of a deadline task should be meaningful. :-(");
+            System.out.println(INDENTATION + "Please enter according to the format eg.description /by YYYY-MM-DD");
+            break;
+        case "error_event_at":
+            System.out.println(INDENTATION + "OOPS!!! The period of a event task should be meaningful. :-(");
+            System.out.println(INDENTATION + "Please enter according to the format eg.description /at YYYY-MM-DD");
+            break;
+        case "find_empty":
+            System.out.println(INDENTATION + "OOPS!!! The enter the keyword of task. :-(");
+        }
```
kouyk: Will it be better to have a `default` branch, since String can take on any arbitrary values and no checks are in place to ensure correctness.

src/main/java/duck/operation/Ui.java
```diff
@@ -0,0 +1,156 @@
...
+        System.out.println(INDENTATION + HORIZON);
+    }
+
+    /**
+     * show the reply for done command
+     * @param number the number of task in task list
```
kouyk: Do you mean `index` of the task?

src/main/java/duck/operation/Ui.java
```diff
@@ -0,0 +1,156 @@
...
+     * @param tasks task list
+     */
+    public void showDateReply(int number, TaskList tasks) {
+        System.out.println(INDENTATION + HORIZON);
+        System.out.println(tasks.getTask(number).getPeriodDays());
+        System.out.println(INDENTATION + HORIZON);
+    }
+
+    public void showFindReply(String[] result) {
+        System.out.println(INDENTATION + HORIZON);
```
kouyk: I see that this wraps around every reply, is it better extract this out to avoid multiple statements like this?

src/main/java/duck/task/Deadline.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * initialize Deadline object
+     * @param description the description of task
+     * @param by          the deadline of task (YYYY-MM-DD)
+     */
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = LocalDate.parse(by);
+        ;
```
kouyk: Extraneous semicolon!

src/main/java/duck/task/Deadline.java
```diff
@@ -0,0 +1,50 @@
...
+     * @return a string show information of task
+     */
+    @Override
+    public String getTaskInfo() {
+        return "[D]" + super.getTaskInfo() + " (by: "
+                + by.getDayOfWeek() + "," + by.format(DateTimeFormatter.ofPattern("MMM d yyyy")) + ")";
+    }
+
+    /**
+     * get period from now to deadline
```
kouyk, line 30: Period is a confusing way to describe the number of remaining days, perhaps `getDaysRemaining` is a better choice?

src/main/java/duck/task/Task.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * mark task as done
+     */
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    /**
```
kouyk, line 60: I assume this method can be made abstract?

src/main/java/duck/task/TaskList.java
```diff
@@ -0,0 +1,77 @@
+package duck.task;
+
+import java.util.ArrayList;
+
+public class TaskList {
+    private ArrayList<Task> taskList = new ArrayList<Task>();
+
+    /**
+     * initialize  the TaskLisk object
+     * @param data use data to form task list
```
kouyk: `data` is very vague and not descriptive of the content of this array, maybe `entries`?

kouyk (commented):
Generally there are no major issues with the code quality. Most of the logic are clear with no deeply nested code. There is sufficient amount of abstraction, it would be even better if some parts are not repeated as often. Naming can also be improved to avoid the awkward use of "of" in variable names and to make it more straightforward to understand what the variable represents. Otherwise the code is rather simple and concise.

src/main/java/duck/operation/Command.java
```diff
@@ -0,0 +1,148 @@
...
+     * @throws IOException
+     */
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws IOException {
+        String[] descriptionSplit;
+        switch (command) {
+        case "bye":
+        case "list":
+        case "hello":
+            ui.showCommandReply(command, tasks);
+            break;
```
jasaaanlim: I like how you let the switch cases fall through for commands requiring the same output. Something most people forget about.

src/main/java/duck/operation/Command.java
```diff
@@ -0,0 +1,148 @@
...
+
+import java.io.IOException;
+import java.time.format.DateTimeParseException;
+
+public class Command {
+    private String command;
+    private String description;
+
+    /**
+     * Initialize the Command object
```
jasaaanlim: Would it be better to leave an empty line between the description and parameter section of the Javadocs (increases readability too)? The description should also begin with 'Initialises'. Just to be in line with the module's conventions.

src/main/java/duck/operation/Gui.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /**
+     * show the reply for done command
+     *
+     * @param number the number of task in task list
+     * @param tasks  task list
+     */
+    public String showDoneReply(int number, TaskList tasks) {
+
```
jasaaanlim: Perhaps these additional newlines should be removed to stay in in line with the coding standards. I noticed the same for a few methods as well.

src/main/java/duck/task/TaskList.java
```diff
@@ -0,0 +1,77 @@
...
+     */
+    public TaskList(String[] data) {
+        for (int i = 0; data[i] != null; i++) {
+            String datas = data[i];
```
jasaaanlim: Would it be better to use a more intuitive variable name e.g. tasks?

src/main/java/duck/task/TaskList.java
```diff
@@ -0,0 +1,77 @@
...
+            String[] dataSplit = datas.split("\\|");
+            switch (dataSplit[0].trim()) {
+            case "T":
+                taskList.add(new Todo(dataSplit[2].trim()));
+                break;
+            case "D":
+                taskList.add(new Deadline(dataSplit[2].trim(), dataSplit[3].trim()));
```
jasaaanlim, line 23: Perhaps for greater readability for future collaborations or debugging, the dataSplit could be assigned to multiple variables with a more intuitive name? E.g. taskDescription, taskDeadline

src/main/java/duck/Duke.java
```diff
@@ -0,0 +1,142 @@
...
+import javafx.scene.image.Image;
+import javafx.scene.image.ImageView;
+import javafx.scene.layout.AnchorPane;
+import javafx.scene.layout.Region;
+import javafx.scene.layout.VBox;
+import javafx.stage.Stage;
+
+import java.io.IOException;
+
+public class Duke extends Application{
```
jasaaanlim: Would it be more consistent to rename your Duke class to Duck just to be more consistent with your project name?

src/main/java/duck/operation/Ui.java
jasaaanlim: I agree with @kouyk that the `HORIZON` variable could be renamed to something more appropriate and intuitive. You may also want to consider setting it to final since there is no intention of modifying the static string anymore.

jasaaanlim (commented):
Looks good! Just some minor coding standard issues.

PR #98, opened 2021-01-24
[Soh Jun Qi] iP

sjq-sohjunqi:


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,209 @@
...
+    * <li>todo <String:name> - Creates an item type of "todo" inside TaskList</li>
+    * <li>deadline <String:name> /by <String:date> - Creates an item type of "deadline" inside TaskList. The date should be specified in an YYYY-MM-DD format</li>
+    * <li>event <String:name> /at <String:date> - Creates an item type of "event" inside TaskList. The date should be specified in an YYYY-MM-DD format</li>
+    * <li>delete <int:number> - Deletes item number specified (1-based)</li>
+    * </ul>
+    * @param in the input command entered by user
+    * @TaskList the TaskList object the commands will be processed on
+    * @return String object containing results of processing 
+    */
+    public String process(String in, TaskList tl){
```
glatiuden: Maybe the method name can be changed slightly to explain the method more intuitively? E.g. processInputToAction? There should be a space between `)` and `{`. I noticed this on your other methods and some of the if statements as well.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,209 @@
...
+        
+        try {
+
+            String res = "";
+            
+            String[] split = in.split("\\s");
```
glatiuden, line 47: Perhaps a more meaningful name can be given to this variable in order to avoid confusion, and preferably in plural form since it is a String collection? E.g. splitStrings

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,209 @@
+package duke;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeParseException;
+import java.time.format.DateTimeFormatter;
+
+public class Parser {
+    private boolean end;
```
glatiuden: Would it be better if it was `isEnd`? For making a boolean variable "sounds like" boolean. 
```suggestion
    private boolean isEnd;
```

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,128 @@
...
+    * @return Properties of the Task
+    */
+    public String toString() {
+        
+        String str = "["+type+"]";
+        
+        if (done)
+            str += "[X] "+name;
+        else
+            str += "[ ] "+name;
```
glatiuden: Java coding standard stated that: `Single statement conditionals should still be wrapped by curly brackets`. Perhaps you could tweak it a little bit? I noticed the same issue on some of the other if-else statements as well.

glatiuden (commented):
LGMT. Just a few comments on the code quality guidelines.

src/main/java/duke/DialogBox.java
```diff
@@ -0,0 +1,20 @@
...
+public class DialogBox extends HBox {
+    private Label text;
+    private ImageView displayPicture;
+
+    public DialogBox(Label l, ImageView iv) {
+        text = l;
+        displayPicture = iv;
+
+        text.setWrapText(true);
+
```
Cp-John, line 17: Are the blank lines here necessary?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,190 @@
...
+        dialogContainer.getChildren().add(getDialogLabel("Hello from\n" + logo));
+    }
+
+    /**
+     * This method initiates the chatbot of the Duke program. It call upon various
+     * classes responsible for different functionalities of Duke.
+     */
+    @Override
+    public void start(Stage stage) {
+        Parser p = new Parser();
```
Cp-John: Why don't you use the full name parser as the variable name? The same for others.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,209 @@
...
+                split = split[1].split("\\s/by\\s");
+                
+                if (split.length <= 1) {
+                    throw new DukeException("OOPS!!! A date and time is needed.");
+                }
+                
+                LocalDate dt = LocalDate.now();
+                
+                try {
+
```
Cp-John: Why do you put a blank line here?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,209 @@
...
+    * <li>event <String:name> /at <String:date> - Creates an item type of "event" inside TaskList. The date should be specified in an YYYY-MM-DD format</li>
+    * <li>delete <int:number> - Deletes item number specified (1-based)</li>
+    * </ul>
+    * @param in the input command entered by user
+    * @TaskList the TaskList object the commands will be processed on
+    * @return String object containing results of processing 
+    */
+    public String process(String in, TaskList tl){
+        
+        try {
```
Cp-John, line 44: Each case block is too long here. Perhaps abstract part of them into methods with reasonable names which increases readability of your code.

src/test/java/duke/DukeTest.java
```diff
@@ -0,0 +1,82 @@
+package duke;
+
+import duke.*;
```
Cp-John, line 3: Remember to always import classes explicitly according to Java coding standard.

Cp-John (commented):
LGTM. But there are several unnecessary blank lines.

PR #99, opened 2021-01-24
[Nicholas Tanvis] iP

nicholastanvis:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,39 @@
...
+        name = n;
+        timestamp = LocalDate.parse(t);
+    }
+
+    /**
+     * Returns the string representation of the task for storage purposes.
+     * @return the string representation
+     */
+    public String toText() {
+        String d = isDone ? "+" : "-";
```
eksinyue, line 25: Maybe it would be good to use named constants for "+" and "-" to avoid the use of magic strings in code?

src/main/java/Parser.java
```diff
@@ -0,0 +1,41 @@
...
+            return new Command(CommandType.ADD_TODO, xs);
+        } else if (command.equals("deadline")) {
+            String xs = String.join(" ", args);
+            args = xs.split(" /by ");
+            return new Command(CommandType.ADD_DEADLINE, args[0], args[1]);
+        } else if (command.equals("event")) {
+            String xs = String.join(" ", args);
+            args = xs.split(" /at ");
+            return new Command(CommandType.ADD_EVENT, args[0], args[1]);
+        } else if (command.equals("delete")) {
```
eksinyue, line 36: Maybe named constants can be used for command tyles and delimiters too to avoid having magic strings.

src/main/java/Ui.java
```diff
@@ -0,0 +1,56 @@
...
+
+/** Describes the wrapper class for handling UI. */
+public class Ui {
+    public final String INVALID_ARGUMENT = "Sorry, you have supplied invalid arguments.";
+    public final String INVALID_COMMAND = "Sorry, I don't understand the command.";
+    public final String ADD_TASK = "I have added the task to the task list!";
+    public final String SHOW_TASKS = "Here are your tasks!";
+    public final String COMPLETE_TASK = "I have marked that task as done!";
+    public final String DELETE_TASK = "I have deleted that task from the task list!";
+    public final String FIND_TASKS = "Finding tasks that match your search keyword...";
```
eksinyue, line 11: I like how you named messages for different scenarios, which improves the readability of the code.

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
...
+            tasks = new TaskList(new ArrayList<>());
+        }
+    }
+
+    /**
+     * Executes the given command.
+     *
+     * @param c The command object
+     */
+    public void executeCommand(Command c) {
```
eksinyue: I personally think this method can be slightly too long and has too many responsibilities. In order to follow SLAP closely, maybe you can consider extract the methods to check invalid inputs, perform the task, and print feedback for the task.

eksinyue (commented):
Overall, I think you did a good job following the Java coding standard, and your code is organized and easy to read. Just brought up some points which I think can be further improved. Hope you find them helpful!! ◡̈

src/main/java/Parser.java
nicholastanvis: does the CommandType enum not suffice?

src/main/java/Task.java
```diff
@@ -0,0 +1,14 @@
+/** Describes generic Task objects. */
+abstract public class Task {
+    protected boolean isDone;
```
Jacob-Pang, line 3: Would it be better to set default value false for Task so that the subclasses do not require setting it in their respective constructors?

src/main/java/Ui.java
```diff
@@ -0,0 +1,56 @@
...
+     */
+    public void echo(String message) {
+        System.out.println(name + ": " + message);
+    }
+
+    /**
+     * Greets by echoing the greeting message.
+     */
+    public void greeting() {
+        this.echo("Hi, I am " + name + "! How can I help you?");
```
Jacob-Pang: Is the syntax "this" necessary for calling the echo method?

nicholastanvis: i believe so

src/main/java/Task.java
nicholastanvis: good idea, how did i miss that?

src/main/java/Storage.java
```diff
@@ -0,0 +1,67 @@
...
+import java.io.PrintWriter;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class Storage {
+    public static String DEFAULT_PATH = "./data.txt";
+
+    public static void ensureFileExists() {
+        File file = new File(DEFAULT_PATH);
+        if (file.exists()) return;
```
jay9645: Hi, according to the module's coding standard, single statement conditionals should still be wrapped by curly brackets

src/main/java/Storage.java
```diff
@@ -0,0 +1,67 @@
...
+        try {
+            PrintWriter writer = new PrintWriter(DEFAULT_PATH);
+            for (Task t : xs) {
+                writer.println(t.toText());
+            }
+
+            writer.close();
+        } catch (FileNotFoundException e) {
+
+        }
```
jay9645: I was wondering if it is bad practice to leave the catch block empty? Maybe you can consider adding a statement to print the error message.

src/main/java/Command.java
```diff
@@ -0,0 +1,24 @@
+import java.util.ArrayList;
+
+public class Command {
+    public CommandType type;
+    public ArrayList<String> args;
```
jay9645, line 6: Is it possible to make this protected? The coding standard says class variables should not be public

jay9645 (commented):
Overall the code was really good and easy to read!

PR #102, opened 2021-01-24
[Lee Hong Thern James] iP

jamesleeht:


hojiefeng:
Overall, the code does not have any obvious code standard violations.

src/main/java/Parser.java
```diff
@@ -0,0 +1,43 @@
...
+
+    TaskList taskList;
+    Ui ui;
+
+    public Parser(TaskList taskList, Ui ui) {
+        this.taskList = taskList;
+        this.ui = ui;
+    }
+
+    public boolean parse(String input) throws DukeException {
```
rhgndf: Would it be better if the input were to be split into two types? One that adds tasks, and the other type does not. It would reduce the total line count of the function, making it easier to read.

rhgndf (commented):
Follows coding standards well

src/main/java/Todo.java
```diff
@@ -1,11 +1,20 @@
...
+        if (!body.trim().isEmpty()) {
+            return new Todo(body);
+        } else {
+            throw new DukeException("Todo description cannot be empty");
+        }
+    }
+
+    @Override
+    public String toString() {
+        return "[T]" + getStatusIcon() + " " + desc;
```
jared98lyj, line 44: Can store the Task type as an attribute so you dun hard code, e.g., taskType = "T"

src/main/java/Event.java
```diff
@@ -7,7 +7,10 @@ public Event(String desc, String at) {
     }
 
     static public Event parse(String input) throws DukeException {
-        String body = input.split("event")[1].strip();
+        String[] arr = input.split("event");
+        if (arr.length < 2) throw new DukeException("Event description cannot be empty");
```
jared98lyj: Try catch block?

src/main/java/Storage.java
```diff
@@ -0,0 +1,100 @@
...
+            List<String> lines = Files.readAllLines(Paths.get(path));
+            return lines;
+        } catch (IOException e) {
+            e.printStackTrace();
+            return null;
+        }
+    }
+
+    private Task parseLine(String line) throws DukeException {
+        String[] parts = line.split("\\|");
```
pavz02, line 68: Perhaps the instead of 'parts' the array name could be more descriptive.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,95 @@
...
+        return new TaskList(result);
+    }
+
+    /**
+     * Get specific task by index.
+     *
+     * @param index Index of task
+     * @return Task at that index
+     */
+    public Task get(int index) {
```
pavz02, line 45: This method could have been called getTask instead for more clarity.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,95 @@
...
+        return selected;
+    }
+
+    /**
+     * Delete task.
+     *
+     * @param index Index of task
+     * @return Deleted task.
+     */
+    public Task delete(int index) {
```
pavz02, line 79: Similar comments for this delete method and the add method below. I think deleteTask and addTask could be better method names.

src/main/java/Parser.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Parses user input into commands that are executed.
+     *
+     * @param input User input.
+     * @return flag to stop the bot
+     * @throws DukeException If input parsing fails.
+     */
+    public String parse(String input) {
+        try {
```
pavz02, line 25: Perhaps switch-case could be used instead of if-else statements.

pavz02 (commented):
Overall I think your code was very well written, I just added some minor nitpicks here and there!

PR #105, opened 2021-01-24
[Guan Yang Ze] iP

guanyz:


src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * Finds all Tasks whose descriptions contain the search parameter and returns them in a TaskList.
+     * @param search Search parameter, as a String.
+     * @return TaskList containing all the matching Tasks.
+     */
+    public TaskList find(String search) {
+        TaskList matchingTasks = new TaskList();
+        String searchPattern = "^.*" + search.toLowerCase() + ".*$";
```
nicholasnge: ^*!_ &^%%**$$$$

nicholasnge (commented):
not bad, id hire you

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,38 @@
...
+public abstract class Task {
+    protected final String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false; // tasks always start as not done
+    }
+
+    public boolean isDone() { return isDone; }
```
GJ0407790: I think the method body can put in a separate line so that it is more consistent with other method style.

GJ0407790 (commented):
Overall, the code looks tidy and follows the coding standard.

1. But I think for some part in the method body, can use a new line to separate blocks that are relevant to improve readability. This is done in some classes but not in some other classes.
2. Can add a newline at the end of each class. 

Good job for following the coding standard.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Loads list of Tasks from the text file specified in the Storage object's filepath.
+     *
+     * @return ArrayList of the Tasks stored in the save data file.
+     * @throws DukeException If an Exception occurs as a result of the stored save data being malformed.
+     */
+    public ArrayList<Task> load() throws DukeException {
+        ArrayList<Task> tasks = new ArrayList<>();
+
+        try {
```
pavz02, line 28: Perhaps you could try to decrease the number of nested blocks.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+            File savefile = new File(filepath);
+            Scanner saveReader = new Scanner(savefile);
+
+            while (saveReader.hasNextLine()) {
+                String savedata = saveReader.nextLine();
```
pavz02: I believe that savedata does not follow the coding standards, and it should be saveData instead.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+
+            StringBuilder saveStringBuilder = new StringBuilder();
+            for (int i = 1; i <= tasks.size(); i++) {
+                saveStringBuilder.append(tasks.get(i).getSaveString());
+            }
+
+            saveWriter.write(saveStringBuilder.toString());
+            saveWriter.close();
+        } catch (Exception e) {
+            String errorMsg = "Save file not found!\n" +
```
pavz02: I think there is a slight inconsistency with the coding standards in errorMsg and that the '+' should be in the second line in front of the "Please don't manually edit the save file.";

src/main/java/duke/TaskParser.java
```diff
@@ -0,0 +1,85 @@
...
+    private static final int SPLIT_LIMIT = 2;
+
+    /**
+     * Parses a String representation of a Task into a Task object, and returns the resultant Task object.
+     *
+     * @param taskString String representation of the Task to be parsed.
+     * @return Task object parsed from the input String.
+     * @throws DukeException If an Exception occurs as a result of the String being malformed.
+     */
+    public static Task parseTask(String taskString) throws DukeException {
```
pavz02: Similar to my above comment, I think you could try to reduce the amounts of nesting in the method parseTask.

pavz02 (commented):
Overall I think your code is good, so I added a few nitpicks here and there!

PR #108, opened 2021-01-24
[Noel Mathew Isaac] iP

noelmathewisaac:


src/main/java/Duke.java
```diff
@@ -1,10 +1,57 @@
+import java.io.FileNotFoundException;
+import java.util.Scanner;
+
+
+/**
+ * A personal task managing chatbot project.
+ */
 public class Duke {
+    private TaskList tasks;
```
chewterence, line 7: I like how your variables are named intuitively 👍

src/main/java/Parser.java
```diff
@@ -0,0 +1,90 @@
...
+     */
+    public static void parseInput(String input, TaskList tasks, Storage storage) {
+        try {
+            if (input.equals("list")) {
+                tasks.listTasks();
+            } else if (input.startsWith("done")) {
+                int index = Integer.parseInt(input.replaceAll("[^-0-9]", ""));
+                if (index > tasks.size() || index <= 0) {
+                    throw new DukeException("The list item number provided is invalid");
+                }
```
chewterence: Perhaps nesting if-else statements like this might be a little confusing? Maybe consider using a switch case?

src/main/java/Parser.java
```diff
@@ -0,0 +1,90 @@
...
+                storage.writeTasksToFile(tasks.getTaskList());
+            } else if (input.startsWith("todo")) {
+                String description = input.replace("todo", "");
+                ToDo todo = new ToDo(description);
+                if (description.strip().equals("")) {
+                    throw new DukeException("todo description cannot be empty");
+                }
+                tasks.addTask(todo);
+                tasks.printAddedTask(todo);
+                storage.writeTasksToFile(tasks.getTaskList());
```
chewterence: Should the return result of tasks.getTaskList() be directly passed into writeTasksToFile()?
Perhaps it could be neater to save the variable and pass the variable into the second function, thus sticking to the same level of abstraction.
Maybe:
TaskList t = tasks.getTaskList();
storage.writeTasksToFile(t);

src/main/java/Parser.java
```diff
@@ -0,0 +1,90 @@
...
+                    throw new DukeException("deadline description cannot be empty");
+                }
+                if (!input.contains("/by")) {
+                    throw new DukeException("Wrong command format: Missing `/by`");
+                }
+                if (input.split("/by").length < 2) {
+                    throw new DukeException("deadline date/time cannot be empty");
+                }
+                String description = input.split("/by", 2)[0].replace("deadline", "");
+                LocalDate date = LocalDate.parse(input.split("/by", 2)[1].strip());
```
chewterence, line 32: I like how this snippet of code is efficiently condensed into a single line. However, perhaps you could expand it into its parts of split, obtaining the value from the array, then applying .strip()?
I noticed this idea could be applied to other parts of your code too.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,128 @@
...
+     * @return An Integer representing the length of the tasks list.
+     */
+    public int size() {
+        return this.tasks.size();
+    }
+
+    /**
+     * Prints all the tasks in the task list.
+     */
+    public void listTasks() {
```
chewterence: Perhaps a better name for this function would be printTasks()?

src/main/java/TaskList.java
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * Searches for tasks based on a keyword.
+     *
+     * @param keyword A String indicating the keyword to search for.
+     */
+    public void findTask(String keyword) {
+        ArrayList<Task> filteredTasks = new ArrayList<>();
+        for (Task task : this.tasks) {
+            if (task.getDescription().toLowerCase().contains(keyword.toLowerCase())) {
```
chewterence: Perhaps extract out the statement within the parenthesis for easier reading?

chewterence (commented):
Overall your code is really easy to read and largely adheres to the coding quality conventions, apart the suggestions i have highlighted.

PR #111, opened 2021-01-25
[Keith Low Jun] iP

fairyinabottle4:


src/main/java/Duke.java
```diff
@@ -1,10 +1,237 @@
...
+        ui = new Ui();
+        storage = new Storage(filepath);
+        tasks = new TaskList(storage.load());
+        name = "Link";
+    }
+
+
+    @Override
+    public void start(Stage stage) throws Exception {
+    //Step 1. Setting up required components
```
ivantjh: Comment indentation should be aligned to method code.

src/main/java/Duke.java
```diff
@@ -1,10 +1,237 @@
...
+
+        return textToAdd;
+    }
+
+
+    /**
+     * You should have your own function to generate a response to user input.
+     * Replace this stub with your completed method.
+     */
+    String getResponse(String input, String[] parts) throws InsufficientArgumentsException, FileNotFoundException {
```
ivantjh: Pretty long method. Could be good to abstract out the logic for individual case clauses.

src/main/java/Parser.java
```diff
@@ -0,0 +1,45 @@
+/**
+ * Parser helps to extract the information in txt files stored in strings into
+ * a desired format.
+ */
+public class Parser {
+
+    /**
+     * Takes in a line from the txt file and outputs the description of a Todo object.
+     * @param str The line read in from the txt file.
```
ivantjh, line 10: Should have an empty line between description and parameters.

src/main/java/Storage.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * Reads the existing txt file and stores the text into Task objects.
+     * If the file does not exist, it creates a new file.
+     * @return the list of tasks that contains the task objects.
+     */
+    public ArrayList<Task> load() {
+        try {
+            File myFile = new File("duke.txt");
+            if (myFile.createNewFile()) {
```
ivantjh: May be better to return `tasks` here to avoid using the else clause at the bottom. This will avoid further nesting of code.

ivantjh: e.g.
```java
if (myFile.createNewFile()) {
    System.out.println(...);
    return tasks
}

System.out.println("File already exists");
...
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,237 @@
...
+                for (int i = 1; i < parts.length; i++) {
+                    str.append(" ");
+                    str.append(parts[i]);
+                }
+                String taskString = str.toString();
+                Todo todo = new Todo(taskString);
+                tasks.add(todo);
+                storage.save();
+                return todo.toString();
+            } else {
```
ivantjh: Could have line breaks in between to make it more readable for others. More info [here](https://nus-cs2103-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#structure-code-logically).

src/main/java/Duke.java
ivantjh: Or put them into their own `Command` classes

src/main/java/Storage.java
```diff
@@ -0,0 +1,92 @@
...
+                System.out.println("File created: " + myFile.getName());
+            } else {
+                System.out.println("File already exists");
+                Scanner myReader = new Scanner(myFile);
+                while (myReader.hasNextLine()) {
+                    String type = myReader.nextLine();
+                    if (type.contains("[T]")) {
```
ivantjh: Would be good if the Tasks themselves could parse the string to create back the instance, e.g.

```java
if (type.contains("[T]")) {
    Task todo = ToDo.createFromStoredData(type);
} else if (type.contains("[D]") {
    Task todo = Deadline.createFromStoredData(type);
} else if ... {
    ...
}
```

ivantjh: So the code would be more readable :)

ivantjh (commented):
No major code quality issues 😄 , try to delegate the class-specific (e.g. Commands and Tasks) logic to their individual classes. It will make the code more readable. 👍

src/main/java/Storage.java
fairyinabottle4: Noted. Thank you for the feedback.

src/main/java/Storage.java
fairyinabottle4: Noted. Thank you for the feedback.

src/main/java/Parser.java
fairyinabottle4: Noted. Thank you for the feedback.

src/main/java/Duke.java
fairyinabottle4: Noted. Thank you for the feedback.

src/main/java/Duke.java
fairyinabottle4: Noted. Thank you for the feedback.

src/main/java/Duke.java
fairyinabottle4: Noted. Thank you for the feedback.

PR #114, opened 2021-01-25
[Donavan Lim] iP

Donavanty:


src/main/java/kobe/Commands.java
```diff
@@ -0,0 +1,37 @@
+package kobe;
+
+import java.lang.System;
+
+public class Commands {
+    public static String ind = "    ";
```
Eriksen2411, line 6: I think you want to put "ind" here as indentation. But i think it should be more expressive because it's attribute. Can be "indentation".

Eriksen2411, line 7: I think you want to put "ind" here as indentation. But i think it should be more expressive because it's attribute. Can be "indentation".

src/main/java/kobe/Commands.java
```diff
@@ -0,0 +1,37 @@
...
+    public static String line = ind + "____________________________________________________________\n" + ind;
```
Eriksen2411, line 8: Also, i do not think that Command class is something that deals with Ui process, cause for this class you are using indentation and line as its attributes, which sounds not like a Command. Just my opinion.

src/main/java/kobe/CustomExceptions.java
```diff
@@ -0,0 +1,29 @@
+package kobe;
+
+class CustomExceptions {
```
Eriksen2411, line 3: Plural form also

src/main/java/kobe/KobeN.java
```diff
@@ -0,0 +1,58 @@
+package kobe;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Scanner;
+
+public class KobeN {
+    private Storage storage;
+    private TaskList tasks; //ArrayList<Tasks> tasks
```
Eriksen2411, line 23: attribute name should be "taskList" to demonstrate correctly what it means for. Tasks are separate tasks that maybe not related to each other, while taskList is different.

src/main/java/kobe/Task.java
```diff
@@ -0,0 +1,138 @@
...
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+class Task {
+    private boolean done;
+    private String taskName;
+    private String type;
+    private String condition;
+    private LocalDate date;
+    private boolean isConditionLocalDate;
```
Eriksen2411, line 12: this attribute name "isConditionLocalDate" is not expressive i think. As i read this, I cannot tell what is the meaning of this attribute. Is it because of the order of the words.

src/main/java/kobe/Ui.java
```diff
@@ -0,0 +1,47 @@
+package kobe;
+
+public class Ui {
```
Eriksen2411, line 7: Here the usage of indentation and line seems to be more correct and sensible. However, as i recall, it should not be named "line" and "line2". I mean there is number in it. Should be something without number and more expressive.

PR #121, opened 2021-01-25
[Wu Weiming] iP

ming-00:
Week 3 individual project assignment


src/main/java/Duke.java
```diff
@@ -1,10 +1,64 @@
...
 public class Duke {
+    private TaskList taskList;
+    private Storage storage;
+    private Ui ui;
+
+
+    /**
+     * Main method for Duke class.
+     *
+     * @param args
```
jlxw48, line 24: Did you accidentally leave out the description for the parameter?

src/main/java/Task.java
```diff
@@ -0,0 +1,48 @@
+/**
+ * This class represents a task that can be created
+ * through user command.
+ */
+public class Task {
```
jlxw48, line 5: Just my opinion: any reason why you chose to use a normal class intead of an abstract class? I notice that you never instantiate a Task object in your main code anyway.

src/main/java/Task.java
```diff
@@ -0,0 +1,48 @@
+/**
+ * This class represents a task that can be created
```
jlxw48, line 2: Do you think this comment is misleading? Since Task cannot be created but rather Task's children classes.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * Class to handle all the tasks.
+ */
+public class TaskList {
+    private static final String ADD_TODO_COMMAND = "todo";
+    private static final String ADD_DEADLINE_COMMAND = "deadline";
+    private static final String ADD_EVENT_COMMAND = "event";
+
+    List<Task> lst = new ArrayList<>();
```
jlxw48, line 21: Is this `lst` variable meant to be accessed outside this class? If not, perhaps a getter method would be more appropriate rather than giving direct access to the other classes, which would result in undesirable side effects.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,178 @@
...
+    public void iterateList() {
+        System.out.println("Here are the items in your list:");
+        int i = 1;
+        for (Task item : lst) {
+            System.out.println(i + ". " + item.toString());
+            i++;
+        }
+    };
+
+    public List<Task> fetchTasks() {
```
jlxw48, line 39: Any reason for the choice of name for this method? The name suggests that it will return Tasks, but instead if gets the `lst` and returns it. Would it be more appropriate to call it `getXXX` instead? Just some thoughts!

src/main/java/TaskList.java
```diff
@@ -0,0 +1,178 @@
...
+        } catch (IndexOutOfBoundsException exception) {
+            System.err.println("You don't have so many items, dumbass!");
+        }
+    }
+
+    /**
+     * Adds task to list of tasks if it is a valid input.
+     * @param input user input in command line
+     */
+    public void addTask(String input) {
```
jlxw48: Would it be better to split the logic into different methods for better organisation?

src/main/java/TaskList.java
```diff
@@ -0,0 +1,178 @@
...
+        for (Task item : lst) {
+            if (item.description.toLowerCase().contains(str.toLowerCase())) {
+                System.out.println(i + ". " + item.toString());
+                i++;
+            }
+        }
+    }
+
+    /**
+     * Determines if task is Todo, Deadline or Event and handles them accordingly.
```
jlxw48, line 175: Do you think it would be clearer to explain that this method is only called when read from storage?

jlxw48 (commented):
Generally looks good! 
Good job on documenting uses of the classes and methods.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,48 @@
...
+ * represent a single item with a deadline that users can add to their list.
+ */
+public class Deadline extends Task {
+    protected LocalDate by;
+
+    /**
+     * Overloaded constructor for Deadline class.
+     * @param description description of task.
+     * @param by deadline of task with deadline.
+     */
```
deyixtan, line 16: Should there be an empty line between description and parameter section?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,48 @@
...
+    public Deadline(String description, LocalDate by) {
+        super(description);
+        this.by = by;
+    }
+
+    /**
+     * Overloaded constructor for Deadline class.
+     * @param isDone whether the task is already marked off as done.
+     * @param description description of task.
+     * @param by deadline of task with deadline.
```
deyixtan, line 26: Should there be an empty line between description and parameter section?

src/main/java/Duke.java
```diff
@@ -1,10 +1,64 @@
...
+ * Duke is a personal assistant chat bot that helps users to
+ * keep track of various things.
+ * @author Damith C. Rajapakse, Wu Weiming
+ */
 public class Duke {
+    private TaskList taskList;
+    private Storage storage;
+    private Ui ui;
+
+
```
deyixtan, line 20: Is this additional blank line necessary?

src/main/java/Event.java
```diff
@@ -0,0 +1,50 @@
...
+     *
+     * @return record of event to be saved in file.
+     */
+    @Override
+    public String getDescription() {
+        return "Event`" + this.isDone + "`" + this.description + "`" + this.at;
+    }
+
+    @Override
+    public String toString() {
```
deyixtan, line 46: Perhaps write a descriptive header comments for this method?

src/main/java/Task.java
deyixtan: I agree with @jlxw48

src/main/java/Task.java
```diff
@@ -0,0 +1,48 @@
...
+     * Constructs a Task object.
+     *
+     * @param description information on task to be done.
+     */
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public Task(String description, boolean isDone) {
```
deyixtan, line 19: Perhaps write a descriptive header comments for this method?

src/main/java/Task.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    /**
+     * Marks a task as completed.
+     */
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    public String getDescription() {
```
deyixtan, line 40: Perhaps write a descriptive header comments for this method?

src/main/java/Task.java
```diff
@@ -0,0 +1,48 @@
...
+        return this.isDone + "`" + this.description;
+    }
+
+    @Override
+    public String toString() {
```
deyixtan, line 45: Perhaps write a descriptive header comments for this method?

src/main/java/TaskList.java
deyixtan: Perhaps it could be named as `displayList`?

src/main/java/TaskList.java
```diff
@@ -0,0 +1,178 @@
...
+            System.err.println("Oof, did you type a valid number or not?");
+        } catch (IndexOutOfBoundsException exception) {
+            System.err.println("You don't have so many items, dumbass!");
+        }
+    }
+
+    /**
+     * Adds task to list of tasks if it is a valid input.
+     * @param input user input in command line
+     */
```
deyixtan, line 65: Should there be an empty line between description and parameter section?

src/main/java/TaskList.java
deyixtan: I agree with @jlxw48, perhaps you could further abstract this method?

src/main/java/TaskList.java
```diff
@@ -0,0 +1,178 @@
...
+        } catch (DateTimeParseException e) {
+            System.err.println("Your date formatting is invalid, use YYYY-MM-DD please...");
+            Ui.formatText();
+        }
+    }
+
+    /**
+     * Deletes task from list of tasks.
+     * @param str task number given one-based indexing
+     */
```
deyixtan, line 147: Should there be an empty line between description and parameter section?

src/main/java/Ui.java
```diff
@@ -0,0 +1,31 @@
...
+     * the exit command.
+     */
+    public static void sayGoodbye() {
+        System.out.println("Bye, see you soon! Don't miss me too much.");
+    }
+
+    /**
+     * Prints a horizontal line to format text.
+     */
+    public static void formatText() {
```
deyixtan, line 44: I think it doesn't really format text per se. Maybe naming it something like `printSeparator()` would be better?

src/main/java/Parser.java
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Takes in user input and deconstructs the command given by the user.
+     *
+     * @param input input provided by user on console.
+     * @param taskList list of current tasks.
+     * @return true if the user has executed an exit command.
+     */
+    public boolean parse(String input, TaskList taskList) {
+        if (input.equals(EXIT_COMMAND)) {
```
deyixtan: I think this method should be named to sound like a boolean?

deyixtan (commented):
I think majority of your code complies with the coding standard, except for some nits to fix. LGTM.

PR #130, opened 2021-01-25
[Jonathan Chiang] iP

banchiang:


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,64 @@
...
+     * Parse the user input for Duke.
+     * ALlows duke to identify which command user inputted.
+     * Returns the unique command.
+     *
+     * @param input user input for the command for Duke
+     * @return command of a type matching to the input from user.
+     * @throws DukeWrongCommandException when command is unknown.
+     */
+    public static Command parseInput(String input) throws DukeWrongCommandException {
+        String[] parsedInput = input.split(" ", 2);
```
TeoHoeKeat, line 56: should there be a variable name that is more intuitive at indicating that the variable is a multi-variable?

src/main/java/duke/AddCommand.java
```diff
@@ -0,0 +1,42 @@
...
+     * Task is then added to the task list and stored on disk.
+     *
+     * @param tl task list.
+     * @param ui user interface object.
+     * @param storage storage of task list.
+     * @throws DukeException if task cannot be created.
+     */
+    @Override
+    public void execute(TaskList tl, Ui ui, Storage storage) throws DukeException {
+        Task t = null;
```
TeoHoeKeat, line 33: The coding standard suggests that the Iterator variable (t) should only be used in nested loops only, 
use Nouns for Things

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,78 @@
+package duke;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeParseException;
+
+public class Deadline extends Task {
+
+    protected LocalDateTime by;
```
TeoHoeKeat: Perhaps a more intuitive variable name here?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Scanner;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.time.LocalDateTime;
+
+
+public class Storage {
+    private final static String PATHNAME = "C:/users/chian/Desktop/CS2103/ip/data/";
+    private final static String FILENAME = "data.txt";
```
TeoHoeKeat: Perhaps it better to put "_" between the word as suggested in coding standard for better readability. (PATH_NAME instead of PATHNAME)

TeoHoeKeat (commented):
Peer review for second PR review. mainly to comment on other code quality guideline. 

The code is neat and looks great. I do suggest avoid over-using of abbreviation(t = task, tl = tasklist) for the variable that it is not in a loop as the meaning can be lost for some reader when the method is too long.  Using name to explain is suggest : https://nus-cs2103-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#use-name-to-explain

src/main/java/duke/Deadline.java
banchiang: okay will do thank you :D

src/main/java/duke/Parser.java
banchiang: okay will do thank you :D

src/main/java/duke/Storage.java
banchiang: okay will do thank you :D

src/main/java/duke/AddCommand.java
banchiang: Hi I don't think they explicitly stated t is a iterator variable. Also they mentioned that variables in a small scope can have short names so I think there need not be change!

PR #133, opened 2021-01-25
[Png Zheng Jie, Sebastian] iP

spzj:


src/main/java/Duke.java
```diff
@@ -1,10 +1,222 @@
...
+
+            default:
+                displayPrompts();
+        }
+    }
+
+    // Marks a task as done and informs the user about it
+    private static void completeTask(String promptDescription) throws DukeException {
+        try {
+            int taskNo = Integer.valueOf(promptDescription);
```
NBH99: The name taskNo could imply not having task instead of task number.

NBH99 (commented):
Good job in following the naming standard for code quality.

src/main/java/Duke.java
```diff
@@ -29,10 +29,13 @@ private static void bye() {
...
-                                    + "  1. todo - adds a todo (E.g. todo borrow book)\n"
-                                        + "  2. deadline - adds a deadline (E.g. deadline return book /by Sunday\n"
-                                            + "  3. event - adds an event (E.g. event project meeting /at Mon 2-4pm)\n"
-                                                + "  4. list - displays the list of tasks\n" + "  5. bye - terminates Duke ☹";
+                                + "  1. todo - adds a todo (E.g. todo borrow book)\n"
+                                + "  2. deadline - adds a deadline (E.g. deadline return book /by Sunday\n"
+                                + "  3. event - adds an event (E.g. event project meeting /at Mon 2-4pm)\n"
+                                + "  4. delete - removes a task from the lists of task\n"
+                                + "  5. list - displays the list of tasks\n"
+                                + "  6. bye - terminates Duke ☹";
```
Juzzanoob: I like how you indicate the valid commands to the user!

src/main/java/Duke.java
```diff
@@ -119,14 +122,38 @@ private static void addTask(String promptDescription) throws DukeException {
...
+    }
+
+    // Deletes a task from the taskList
+    private static void deleteTask(String promptDescription) throws DukeException {
+        try {
+            int taskNo = Integer.valueOf(promptDescription);
+            if (taskNo > 0 && taskNo <= tasksList.size()) {
+                counter--;
+                Task taskToDelete = tasksList.get(taskNo - 1);
+                String deletedMessage = "Noted. I've removed this task:\n" + "  " + taskToDelete + "\n"
```
Juzzanoob: Perhaps "deletedMessage" can be changed into "deletedTaskMessage"? Just so that others won't get confused if the deletedMessage variable appears later and they are confused that it refers to a deleted message.

src/main/java/Duke.java
```diff
@@ -1,10 +1,222 @@
...
+        String systemMessage;
+        String description;
+        String period;
+        Task newTask;
+
+        // Checks if string is null or empty or contains only spaces
+        boolean isDescriptionFilled = (promptDescription != null && !promptDescription.trim().isEmpty());
+
+        switch (taskType) {
+            case TODO:
```
fairyinabottle4: The case statement should be at the same indentation level as switch statement

src/main/java/Duke.java
```diff
@@ -1,10 +1,222 @@
...
+                        strings = promptDescription.split(" /at");
+                        description = strings[0];
+                        period = (strings.length > 1 ? strings[1].split(" ", 2)[1] : "");
+
+                        if (period != null && !period.trim().isEmpty()) {
+                            newTask = new Event(description, period);
+                            tasksList.add(newTask);
+                            counter++;
+                            systemMessage = "Got it. I've added this task:\n" + "  " + newTask + "\n" + "Now you have "
+                                                        + counter + (counter <= 1 ? " task" : " tasks") + " in the list.";
```
fairyinabottle4: The indentation is a bit too excessive here, should only be 8 spaces from the line above

fairyinabottle4 (commented):
Overall your code looks good! Two things to watch out for however. One is the switch case statement that I pointed out, the other is the multiline indentation, of which there are a few instances, although I only highlighted one. I suggest you look through your code again and check for the indentation level of such multiline statements.

PR #146, opened 2021-01-25
[Chester How] iP

chesterhow:


src/main/java/Duke.java
```diff
@@ -1,10 +1,186 @@
...
+        }
+    }
+
+    /**
+     * Marks a task as done
+     *
+     * @param command Command containing the index of the task to mark as done.
+     * @throws IOException If an error occurs while writing to the file.
+     */
+    public void done(Command command) throws IOException {
```
SoulUseless: Since this function is an action, would it be better to name this is a way that sounds like an action? (Eg: markAsDone())?

src/main/java/Ui.java
```diff
@@ -0,0 +1,106 @@
...
+        reply(formatLine("Bye. Hope to see you again soon!"));
+    }
+
+    /**
+     * Prints a reply, having added a task.
+     *
+     * @param task Task that was added.
+     * @param numTasks Total number of tasks in task list.
+     */
+    public void addedTaskReply(Task task, int numTasks) {
```
SoulUseless: Since this is a function that does an action, would it be better to name the function as a action? (Eg: printAddedTaskReply) This would also apply to the below Reply functions as well.

SoulUseless (commented):
Please find below my PR review.

src/main/java/Ui.java
```diff
@@ -0,0 +1,88 @@
...
+     * @param numTasks Number of tasks in task list.
+     * @return Reply to having deleted a task.
+     */
+    public static String deleteReply(Task deletedTask, int numTasks) {
+        return formatLine("Noted. I've removed this task:")
+                + formatLine("  " + deletedTask)
+                + formatLine("Now you have " + numTasks + " tasks in the list.");
+    }
+
+    public static String listMatchingTasks(TaskList tasks) {
```
github-amanda: Header comment is missing for this public method.

src/main/java/Task.java
```diff
@@ -0,0 +1,40 @@
...
+     * Returns the string representation of how the task will be stored in the file.
+     *
+     * @return String
+     */
+    public String toFileString() {
+        char done = this.isDone ? 'X' : ' ';
+        return String.format("%c | %s", done, this.description);
+    }
+
+    public String toString() {
```
github-amanda: Header comment is missing for this public method.

src/main/java/Ui.java
```diff
@@ -0,0 +1,88 @@
...
+    public static String list(TaskList tasks) {
+        StringBuilder msg = new StringBuilder(formatLine("Here are the tasks in your list:"));
+
+        for (int i = 0; i < tasks.size(); i++) {
+            msg.append(formatLine((i + 1) + ". " + tasks.getTaskAt(i)));
+        }
+        return msg.toString();
+    }
+
+    public static String exit() {
```
github-amanda, line 35: Header comment is missing for this public method.

src/main/java/Ui.java
github-amanda: Header comment is missing for this public method.

src/main/java/Ui.java
```diff
@@ -0,0 +1,88 @@
...
+     * Formats given string with indentation and line break
+     *
+     * @param line Line to format.
+     * @return Formatted line.
+     */
+    public static String formatLine(String line) {
+        return line + "\n";
+    }
+
+    public static String getGreeting() {
```
github-amanda, line 15: Header comment is missing for this public method.

github-amanda (commented):
Neat code 😄 Just some missing header comments for public methods.

PR #152, opened 2021-01-25
[Wong Choon Wei] iP

chwoozy:


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,164 @@
...
+
+    /**
+     * Parses given input and sets a new current message based on input.
+     *
+     * @param input the input message.
+     * @return integer. 0 represents an terminating entry, 1 represents a continuing entry.
+     * @throws DukeException the duke exception
+     * @throws IOException   the io exception
+     */
+    public int parse(String input) throws DukeException, IOException {
```
IceBear789: This method seems a bit long. Perhaps you can use helper methods to make the code shorter?

src/main/java/exception/DukeException.java
```diff
@@ -0,0 +1,37 @@
+package exception;
```
IceBear789, line 1: I've noticed that you have implemented quite a lot of subclasses that inherit from DukeException. Since the only difference between these subclasses are the error message printed, perhaps they can be combined into one subclass? (Together with a new attribute for displaying the error message)

src/main/java/task/Task.java
```diff
@@ -0,0 +1,60 @@
+package task;
```
IceBear789, line 1: Since any task must be either a todo, a deadline or an event, maybe you can consider making this an abstract class?

src/main/java/duke/DukeStorage.java
```diff
@@ -0,0 +1,165 @@
...
+            case "Event": {
+                currTask = "Event | " + t.getDescription() + " | "
+                        + ((Event) t).getSimpleAt();
+                break;
+            }
+            }
+            content.append(currTask);
+            content.append("\n");
+        }
+        FileWriter fw = new FileWriter(this.fileName);
```
IceBear789, line 163: Personally I don't feel that fw is a very clear acronym. Maybe a clearer variable name here?

src/main/java/duke/DukeTaskList.java
```diff
@@ -0,0 +1,75 @@
...
+     * @param taskList the task list
+     */
+    DukeTaskList(ArrayList<Task> taskList) {
+        this.taskList = taskList;
+    }
+
+    /**
+     * Add new task into task list.
+     *
+     * @param t the t
```
IceBear789: The description for the @param could be clearer. I've noticed the same issue in some other places as well.

IceBear789 (commented):
Some of the methods seem a bit long. You may want to split them into smaller methods for better readability. Other than that, the code is generally quite clean and comments are generally quite clear. With the exception of a few minor coding violations, LGTM.

chwoozy:
> Some of the methods seem a bit long. You may want to split them into smaller methods for better readability. Other than that, the code is generally quite clean and comments are generally quite clear. With the exception of a few minor coding violations, LGTM.

Thank you for your input! I will be sure to make the necessary changes according to your feedback! 😄

PR #156, opened 2021-01-25
[Isaiah Koh] iP

zaiahkoh:


src/main/java/Sweh.java
```diff
@@ -9,8 +23,15 @@
 import static util.Parser.getArgMap;
 import static util.Parser.getCommand;
 
-public class Sweh {
+public class Sweh extends Application {
     public static TaskList taskList = new TaskList();
+    private ScrollPane scrollPane;
+    private VBox dialogContainer;
```
Sharptail: Should there be a JavaDoc Comment to tell use what does this class represent?

src/main/java/Sweh.java
```diff
@@ -9,8 +23,15 @@
...
+    private TextField userInput;
+    private Button sendButton;
+    private Scene scene;
+    private Image user = new Image(this.getClass().getResourceAsStream("/images/DaUser.png"));
+    private Image duke = new Image(this.getClass().getResourceAsStream("/images/DaDuke.png"));
```
Sharptail: I like that you use "user" and "duke" to differentiate the respective images. However, would it be more appropriate to name them as "userImage" and "dukeImage"? As "user" and "duke" might be a more suitable name for the respective Object Class.

src/main/java/nodes/DialogBox.java
```diff
@@ -0,0 +1,46 @@
...
+import javafx.geometry.Pos;
+import javafx.scene.Node;
+import javafx.scene.control.Label;
+import javafx.scene.image.ImageView;
+import javafx.scene.layout.HBox;
+
+public class DialogBox extends HBox {
+    private Label text;
+    private ImageView displayPicture;
+
```
Sharptail: Should there be JavaDoc comment to tell use what the class represents? I noticed the same issue in several other classes too. It would be good to have a JavaDoc comments on the class so that we can know more information about this class.

src/main/java/Sweh.java
```diff
@@ -24,15 +26,16 @@ public static void main(String[] args) {
...
     }
 
     public static String respond(String command) {
+
         switch (command) {
             case "bye":
                 return "Bye. See ya again soon!";
             default:
-                return echo(command);
+                return taskList.addTask(command);
```
Sharptail: Should the switch-case statement have the same indentation?

Sharptail (commented):
Overall great coding standard! It would be helpful to have JavaDoc to tell us about the functionalities of the methods and classes! Good Work!

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,9 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+<<<<<<< HEAD
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-bin.zip
+=======
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.2-bin.zip
+>>>>>>> add-gradle-support
```
daniellau88, line 7: I think you may have forgotten to resolve the conflict here?

gradlew.bat
```diff
@@ -0,0 +1,114 @@
...
+
+@rem Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
+set DEFAULT_JVM_OPTS="-Xmx64m" "-Xms64m"
+
+@rem Find java.exe
+if defined JAVA_HOME goto findJavaFromJavaHome
+
+set JAVA_EXE=java.exe
+%JAVA_EXE% -version >NUL 2>&1
+<<<<<<< HEAD
```
daniellau88, line 43: The conflict here seems to be unresolved too

src/main/java/TaskList.java
```diff
@@ -0,0 +1,160 @@
...
+import static util.Parser.getCommand;
+
+public class TaskList {
+    private final ArrayList<Task> taskList;
+
+    public TaskList() {
+        this.taskList = new ArrayList<>();
+    }
+
+    private String taskCountMsg() {
```
daniellau88: Would `getTaskCountMsg()` be a better name to make it consistent with the other method's namings

src/main/java/util/Formatter.java
```diff
@@ -0,0 +1,30 @@
+package util;
+
+import java.util.List;
+
+public interface Formatter {
```
daniellau88, line 10: Would it be better to implement the class as a concrete class instead? The intention of an interface is to act as a contract.

src/main/java/util/Parser.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    static HashMap<String, String> getArgMap(String input) throws InputMismatchException {
+        HashMap<String, String> argMap = new HashMap<>();
+
+        if (input.split(" ").length < 2) {
+            return argMap;
+        }
+
+        String[] argArr = input.split("/");
```
daniellau88: Maybe you would like to rename it to a more informative name, e.g. argumentsFromInput (might not be the best name, but I can't really think of other alternatives)

src/main/java/util/Parser.java
```diff
@@ -0,0 +1,39 @@
...
+        for (int i = 0; i < argArr.length; i++) {
+            String flag;
+            String param;
```
daniellau88: How about using the pair words key and value, since they will be used in a `HashMap`

src/test/util/StorageTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class StorageTest {
+
+    @org.junit.jupiter.api.Test
```
daniellau88: Should this be `@Test` (looks different from the rest)

daniellau88 (commented):
Overall, code looks good. I added a few reviews on conflicts that you did not resolved (even though they are not coding standards violations, it probably will affect the project). Hopefully, it helps.

PR #158, opened 2021-01-25
[Tan Boon Ji] iP

tanboonji:


src/main/java/ToDoCommand.java
```diff
@@ -0,0 +1,39 @@
...
+        this.description = description;
+    }
+
+    @Override
+    public String execute() throws DukeException {
+        Task newTask = new ToDo(description);
+        taskList.addTask(newTask);
+
+        StringBuilder builder = new StringBuilder();
+        builder.append("Got it. I've added this task:\n\t").append(newTask).append("\nNow you have ")
```
DrWala: It might assist readability to break the `.append` into multiple lines since they are new lines, and it makes it clear what is being appended to the `StringBuilder`

src/main/java/DeadlineCommand.java
```diff
@@ -0,0 +1,53 @@
...
+        this.by = by;
+    }
+
+    @Override
+    public String execute() throws DukeException {
+        Task newTask = new Deadline(description, by);
+        taskList.addTask(newTask);
+
+        StringBuilder builder = new StringBuilder();
+        builder.append("Got it. I've added this task:\n\t").append(newTask).append("\nNow you have ")
```
DrWala: It might assist readability to break the `.append` into multiple lines since they are new lines, and it makes it clear what is being appended to the `StringBuilder`.

This specific style issue exists in a few other places too.

src/main/java/HelpCommand.java
```diff
@@ -0,0 +1,28 @@
+public class HelpCommand extends Command {
+
+    public static final String COMMAND = "help";
+    public static final String ERROR_MESSAGE = "☹ Sorry, please enter a valid command.\n";
+    public static final String COMMAND_LIST = "\tCommands available:\n\t\t" +
```
DrWala: Perhaps the same breaking into multiple lines argument goes here too to improve readability, and ease of maintenance should the order be changed or new commands be added.

DrWala (commented):
LGTM, just a few nits to fix.

src/main/java/com/tanboonji/duke/Duke.java
```diff
@@ -0,0 +1,98 @@
...
+            taskList = storage.load();
+            ui.print("load successful");
+        } catch (DukeException e) {
+            ui.print(e.getMessage());
+        }
+
+        ui.greet();
+    }
+
+    private boolean processInput(String input) {
```
andrea-twl: Perhaps this method could be named in a way that makes it clearer that it returns a boolean?

andrea-twl (approved):
Looks good overall!

PR #159, opened 2021-01-25
[Benedict Chua] iP

BenedictBCJJ:


src/main/java/Duke.java
```diff
@@ -1,10 +1,73 @@
...
+                        break;
+                    case DONE:
+                        Task doneTask = taskList.get(parser.getTaskIndex() - 1);
+                        doneTask.setDone(true);
+                        Ui.execute(Command.DONE, taskList, doneTask.getStatus());
+                        break;
+                    case DEADLINE:
+                    case EVENT:
+                    case TODO:
+                        if (parser.hasDescriptionError()) {
```
lyueyang: Perhaps you could have a helper function to avoid deep nesting for better readability :)

https://nus-cs2103-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-deep-nesting

src/main/java/Storage.java
```diff
@@ -0,0 +1,19 @@
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class Storage {
+    public static void createAndWrite(String path, TaskList listOfTasks) {
+        try {
+            File savedTasks = new File(path);
```
lyueyang: Perhaps savedTask would be better as it's a singular File instead of a List of Files

https://nus-cs2103-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#use-nouns-for-things-and-verbs-for-actions

src/main/java/Duke.java
```diff
@@ -1,10 +1,73 @@
...
+                            Ui.exception(parser.getDescription());
+                        } else if (!taskType.equals(Command.TODO) && parser.hasDateError()) {
+                            Ui.exception(parser.getDate());
+                        }
+                        Task task = taskType.equals(Command.EVENT)
+                                ? new Event(parser.getDescription(), LocalDate.parse(parser.getDate(), formatter))
```
lyueyang: Perhaps you could consider having a helper function here too to make it easier to read. Ternary is honestly great and I get an idea of what you're doing though

https://nus-cs2103-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-complicated-expressions

src/main/java/Event.java
```diff
@@ -0,0 +1,22 @@
...
+    private LocalDate eventTime;
+
+    public Event(String description, LocalDate eventTime) {
+        super(description);
+        this.eventTime = eventTime;
+    }
+
+    @Override
+    public String getStatus() {
+        return "[E]" + super.getStatus() + "(at: " + this.eventTime.format(DateTimeFormatter.ofPattern("MMM d yyyy"))
```
lyueyang: Kind of a long statement here as well, perhaps I would suggest using StringBuilder for better readability :)

src/main/java/Parser.java
```diff
@@ -0,0 +1,67 @@
...
+                case DELETE:
+                case DONE:
+                    this.taskIndex = Integer.parseInt(this.information[1]);
+                    break;
+                case TODO:
+                case DEADLINE:
+                case EVENT:
+                    String[] descriptionAndTime = information.length < 2 ? null : information[1].split("/", 2);
+                    this.description = descriptionAndTime == null ? "DESCRIPTION.ERROR" : descriptionAndTime[0];
+                    this.date = descriptionAndTime == null || descriptionAndTime.length < 2 ? "DATE.ERROR"
```
lyueyang: Perhaps I could suggest declaring a static final variable to define what 2 is supposed to be here, I believe it's to ensure that there are sufficient arguments but maybe you could name it something like MIN_VARS :)

https://nus-cs2103-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-magic-numbers

lyueyang (commented):
Your code was honestly really neat! Just some minor tweaks you could consider :)

src/main/java/Deadlines.java
```diff
@@ -0,0 +1,22 @@
...
+        this.deadlineTime = deadlineTime;
+    }
+
+    @Override
+    public String getStatus() {
+        return "[D]" + super.getStatus() + "(by: " + this.deadlineTime.format(DateTimeFormatter.ofPattern("MMM d yyyy"))
+                + ")";
+    }
+
+    @Override
```
ZechariahTan: Should there be some documentation for this method? I noticed there aren't any javadocs for other methods in the code as well

src/main/java/Duke.java
```diff
@@ -1,10 +1,73 @@
...
+                        Ui.execute(Command.DELETE, taskList, deletedTask.getStatus());
+                        break;
+                    case DONE:
+                        Task doneTask = taskList.get(parser.getTaskIndex() - 1);
+                        doneTask.setDone(true);
+                        Ui.execute(Command.DONE, taskList, doneTask.getStatus());
+                        break;
+                    case DEADLINE:
+                    case EVENT:
+                    case TODO:
```
ZechariahTan: I think a //Fallthrough comment is needed here in case DEADLINE and EVENT according to the coding standard? Similar issue was spotted in Parser and UI.

src/main/java/Parser.java
```diff
@@ -0,0 +1,67 @@
...
+                            : descriptionAndTime[1].split(" ", 2)[1];
+
+                    break;
+                case BYE:
+                case NONE:
+                case LIST:
```
ZechariahTan: Maybe a default case should be added here? I'm not 100% sure but the coding standard provided states that a switch case should have a certain form and it included a default case.

src/main/java/Parser.java
```diff
@@ -0,0 +1,67 @@
...
+                case DONE:
+                    this.taskIndex = Integer.parseInt(this.information[1]);
+                    break;
+                case TODO:
+                case DEADLINE:
+                case EVENT:
+                    String[] descriptionAndTime = information.length < 2 ? null : information[1].split("/", 2);
+                    this.description = descriptionAndTime == null ? "DESCRIPTION.ERROR" : descriptionAndTime[0];
+                    this.date = descriptionAndTime == null || descriptionAndTime.length < 2 ? "DATE.ERROR"
+                            : descriptionAndTime[1].split(" ", 2)[1];
```
ZechariahTan: I like that this ternary was broken into two lines to keep within the 120 char limit but should it be formatted as 

(CONDITION)
? method1()
: method2()

instead?

ZechariahTan (commented):
Looks very clean overall. Just needs some extra comments to describe code functions.

PR #160, opened 2021-01-25
[Ian] iP

IanCKW:


src/main/java/duke/FileReading.java
```diff
@@ -0,0 +1,44 @@
...
+        }
+        return output;
+    }
+
+    /**
+     * Reads file and generates a string of the contents.
+     * Catches FileNotFoundExceptions.
+     *
+     * @param filePath
+     * @return
```
natosy: Missing description of return String?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,76 @@
...
+     * @param duke
+     * @param currLine the current input from the user
+     * @param taskList
+     * @param ui
+     * @param storage
+     * @throws Exception when an invalid command is made. i.e. the first word in the input is
+     *                   invalid
+     * @return
+     */
+    protected String handleCommand(Duke duke, String currLine, TaskList taskList, Ui ui, Storage storage) throws Exception {
```
natosy: Not sure if it's just me, but I feel that this method looks very cluttered. Also, missing return definition? Not sure if it's intentional (or if it's allowed).

natosy (commented):
Barely found any problems! I think your code is pretty organised other than the fact that there was a big chunk where the input was being processed which I found to be cluttered. It doesn't violate any coding guidelines though, just a personal thought!

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
```
XindiTian, line 1: perhaps there should be spacings between the variables and the value?

like distributionBase = GRADLE_USER_HOME and the sorts?

src/main/java/Main.java
```diff
@@ -0,0 +1,46 @@
...
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+    }
+
+    /**
+     * Main Method
+     **/
+
+    public static void main(String[] args) {
```
XindiTian: Does this main function do anything? If not maybe you can delete it to make your code look neater?

XindiTian (commented):
Overall, good job on following the java coding standards! I feel that the only few things to look out for are the spacings for when you assign variables, but these are really minor details that could be easily changed.

Also, it may be because of carelessness but sometimes your javadocs are missing return statements. But this is also not a huge issue and can be solved quickly.

PR #163, opened 2021-01-26
[Lim Li Quan] iP

AhQuanz:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,37 @@
+import java.time.LocalDateTime;
+
+/** Class Deadline that represent a task with a deadline **/
```
douglaswja, line 3: My understanding of the CS2103 Java Coding Standard is that Javadocs should end with `*/` rather than `**/`.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,37 @@
...
+public class Deadline extends Task{
+    /** Symbol to represent a deadline task **/
+    private static final char SYMBOL = 'D';
+    /** deadline of the task **/
+    private LocalDateTime deadline;
+
+    /**
+     * Constructor of a Deadline task.
```
douglaswja, line 11: 2 points:
1. There should be an "Empty line between description and parameter section" - CS2103 Java Coding Standard.
2. There should be "Punctuation behind each parameter description". I.e. Lines 12 and 13 of Deadline.java should end with a period, as shown below.

Consider the following Javadoc:
```
/**
* Constructor of a Deadline task.
* 
* @param desc Description of the deadline task.
* @param deadline Deadline of the deadline task.
*/
```

src/main/java/Event.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /**
+     * Returns the representation of a task.
+     * @return String representation of a task
+     */
+    @Override
+    public String toString() {
+        return String.format("[%c] %s (%s - %s)", SYMBOL, super.toString(),
+                    super.format(this.start), super.format(this.end));
```
douglaswja, line 31: The line break here uses 12 spaces instead of the required 8.

src/main/java/Event.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /**
+     * Returns the representation of a task for saving.
+     * @return String representation of a task for saving
+     */
+    @Override
+    public String save() {
+        return String.format("%c,%s,%s,%s\n", SYMBOL, super.save(),
+                    super.saveFormat(this.start), super.saveFormat(this.end));
```
douglaswja, line 41: The line break here uses 12 spaces instead of the required 8.

src/main/java/Parser.java
```diff
@@ -0,0 +1,166 @@
...
+     * Returns a Command object based on the fullCommand given
+     * @param fullCommand (command entered by the user)
+     * @return Command comamnd
+     * @throws DukeException
+     * @throws DukeDeadlineException
+     */
+    public static Command parse(String fullCommand) throws DukeException , DukeDeadlineException {
+        fullCommand = fullCommand.trim();
+        int firstSpace = fullCommand.indexOf(" ");
+        String keyword = firstSpace == -1 ? fullCommand : fullCommand.substring(0, firstSpace).toLowerCase();
```
douglaswja: The CS2103 Java Coding Standard gives the following examples of acceptable ternary statements:
```
alpha = (aLongBooleanExpression) ? beta : gamma;
alpha = (aLongBooleanExpression)
        ? beta
        : gamma;
```

Consider enclosing the boolean term within parentheses, as in the following:
```
String keyword = (firstSpace == -1) ? fullCommand : fullCommand.substring(0, firstSpace).toLowerCase();
```

douglaswja (commented):
Just a few nits.

src/main/java/Parser.java
```diff
@@ -0,0 +1,166 @@
...
+                }
+                task = new Event(taskArr[2], deadlines[0], deadlines[1]);
+                break;
+            case "D":
+                LocalDateTime deadline = Parser.parseDate(taskArr[3], errorMessage);
+                task = new Deadline(taskArr[2], deadline);
+            }
+            if (taskArr[1] == "1") {
+                task.markAsDone();
+            }
```
douglaswja: Could consider ending the last case statement with a 'break'.

From the CS2103 Java Coding Standard:
The explicit //Fallthrough comment should be included whenever there is a case statement without a break statement.
Rationale: Leaving out the break is a common error, and it must be made clear that it is intentional when it is not there.

douglaswja (commented):
Another nit about a switch statement.

PR #165, opened 2021-01-26
[Umer Siddiqui] iP

umergta:


src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,111 @@
...
+     * @see Event
+     * @see Deadline
+     * @see Paths
+     */
+    public ArrayList<Task> getTaskList() throws IOException {
+        ArrayList<Task> taskList = new ArrayList<>();
+        List<String> fileLines = Files.readAllLines(Paths.get(filePath));
+        for (int i = 0; i < fileLines.size(); i++) {
+            String line = fileLines.get(i);
+            String[] userTask = line.split(" \\| ");
```
internityz, line 65: Distinguish clearly between single-valued and multi-valued variables. Use plural instead

internityz (commented):
LGTM. Just a few nits to fix. :)

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,30 @@
...
+    private LocalDate deadlineDate;
+
+    Deadline(String deadlineDateString, String deadlineDetail) {
+        super(deadlineDetail);
+        try {
+            this.deadlineDate = Task.dateStringToDate(deadlineDateString);
+            this.deadlineDateString = deadlineDate.getMonth().toString().toLowerCase() + " " +
+                    deadlineDate.getDayOfMonth() + " " + deadlineDate.getYear();
+        }
+        catch (DateTimeException e) {
```
kaixiangtay: Do note that the catch statement should start on the same line as the closing braces of try block.

src/main/java/duke/Storage.java
kaixiangtay: To add on what internityz has mentioned above, for the line 53 code in Storage.java, you can consider renaming the Arraylist<Task> taskList to tasks.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,111 @@
...
+                    taskInList = new Event(eventDuration, eventDetail);
+                    taskList.add(taskInList);
+                    break;
+                case ("[D]"):
+                    String deadline = parser.parseDate(userTask[3].split("by: ")[1]);
+                    String deadlineDetail = userTask[2];
+                    taskInList = new Deadline(deadline, deadlineDetail);
+                    taskList.add(taskInList);
+                    break;
+                }
```
kaixiangtay, line 102: One thing I want to highlight here that can be of interest to you is according to the coding standard, each switch statement should include a default statement even if the default does not contain any code.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,97 @@
...
+        System.out.println("Now you have " + (userTaskList.getTaskListSize()) + " tasks in the list.");
+        System.out.println(HORIZONTAL_RULE);
+    }
+
+    public void printFoundTasks(TaskList tasksFound) {
+        System.out.println("Here are the matching tasks in your list:");
+        if (tasksFound.getTaskListSize() == 0) {
+            System.out.println("Sorry. No tasks found :-(");
+        }
+        else {
```
kaixiangtay: Do note that the else statement block should also start on the same line as the closing braces of the if block.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,95 @@
...
+    public TaskList parseFindCommand(String command, TaskList userList) {
+        String keywords = command.substring(5);
+        ArrayList<Task> results = new ArrayList<>();
+        for (Task task : userList.getTaskList()) {
+            if (task.getTaskDetail().contains(keywords)) {
+                results.add(task);
+            }
+        }
+        return new TaskList(results);
+
```
kaixiangtay, line 127: Just a small mistake here of having an extra line of space on line 91 in Parser.java.

PR #167, opened 2021-01-26
[Alzahrani Riyadh Hamdan S] iP

riyadh-h:


src/vergil/components/Parser.java
```diff
@@ -0,0 +1,90 @@
...
+                        CommandType.FIND,
+                        command.split(" ", 2)[1],
+                        null,
+                        0
+                );
+
+            default:
+                throw new VergilException("Unable to resolve command.");
+            }
+         } catch (ArrayIndexOutOfBoundsException | DateTimeParseException e) {
```
JoelHo: Should the indent be 8 spaces instead of 9?

src/vergil/components/Parser.java
```diff
@@ -0,0 +1,90 @@
...
+                        CommandType.TODO,
+                        command.split(" /by ")[0].split(" ", 2)[1],
+                        null,
+                        0);
+
+            case "deadline":
+                return new Command(
+                        CommandType.DEADLINE,
+                        command.split(" /by ")[0].split(" ", 2)[1],
+                        LocalDateTime.parse(command.split(" /by ")[1], DateTimeFormatter.ofPattern("d/M/y HHmm")),
```
JoelHo: Should this be broken up? It may be a bit too complicated

src/vergil/components/Parser.java
```diff
@@ -0,0 +1,90 @@
...
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+public class Parser {
+    public Parser() {
+        // Create a new parser.
+    }
+
+    /**
+     * Parses a given command as String and returns a Command object representation
```
JoelHo, line 25: Should there be a space between the description and params? Seems to occur a lot, but only a minor cosmetic issue

src/vergil/components/TaskList.java
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * Returns the tasks in the list, but as save file entries.
+     * @return a save-file-entries representation of the task list.
+     */
+    public String toSaveString() {
+        StringBuilder str = new StringBuilder();
+
+        for (Task t:tasks) {
+            str.append(t.getSaveString()).append("\n");
```
JoelHo, line 208: Should the appends be broken up? Or perhaps in a newline eg

    str.append(t.getSaveString())
            .append("\n");

Similar as below.

src/vergil/components/Ui.java
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Displays an error message stating the error.
+     * @param message a string describing the actual error.
+     */
+    public void displayError(String message) {
+        System.out.println("Error! " + message);
+    }
+
+    public void displayFindResult(TaskList resultsList) {
```
JoelHo: Should there be a javadoc for this method?

src/vergil/main/Vergil.java
```diff
@@ -0,0 +1,107 @@
+package vergil.main;
+
+import vergil.components.Ui;
+import vergil.components.Storage;
+import vergil.components.Parser;
+import vergil.components.TaskList;
+
```
JoelHo, line 33: Should there be a seperation here? All belong to the vergil package so in my opinion they should be together

src/vergil/main/Vergil.java
```diff
@@ -0,0 +1,107 @@
...
+                    taskList.add(new Event(cmd.getDesc(), cmd.getTime()));
+                    ui.displaySuccess();
+                    storage.save(taskList);
+                    break;
+
+                case FIND:
+                    TaskList resultsList = taskList.find(cmd.getDesc());
+                    ui.displayFindResult(resultsList);
+                    break;
+                }
```
JoelHo: Should there be a default clause to the switch case?

src/vergil/main/Vergil.java
```diff
@@ -0,0 +1,107 @@
...
+import vergil.types.VergilException;
+import vergil.types.Command;
+import vergil.types.Todo;
+import vergil.types.Event;
+import vergil.types.Deadline;
+
+public class Vergil {
+    private Ui ui;
+    private Storage storage;
+    private Parser parser;
```
JoelHo, line 39: Should these fields be final? They are not getting modified (and in fact if they do you might run into issues!) so no harm letting the compiler check for us?

src/vergil/testing/VergilTest.java
```diff
@@ -0,0 +1,36 @@
+package vergil.testing;
```
JoelHo, line 1: Should the tests be in this structure?

src/vergil/types/Command.java
```diff
@@ -0,0 +1,33 @@
...
+        return desc;
+    }
+
+    public LocalDateTime getTime() {
+        return time;
+    }
+
+    public int getListNumber() {
+        return listNumber;
+    }
```
JoelHo: Should there be javadoc comments for these methods?

src/vergil/components/Parser.java
```diff
@@ -0,0 +1,90 @@
...
+
+import vergil.types.Command;
+import vergil.types.CommandType;
+import vergil.types.VergilException;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+public class Parser {
```
JoelHo, line 16: Should there be a javadoc explaining the class? I notice most of your classes are missing this

JoelHo (commented):
Looks good, perhaps there could be slightly better documentation.
Some calls are slightly too long and complex, making them harder to read, maybe split them up into logical chunks instead

src/vergil/components/Parser.java
riyadh-h: Right, I'll have to work on that. Thanks for the reminder!

src/vergil/components/Parser.java
riyadh-h: Yeah, you're right. Nice catch!

src/vergil/components/Parser.java
riyadh-h: Sorry, but can you explain why it looks complicated?

src/vergil/components/Parser.java
riyadh-h: Right, the `catch` statement seems to have an extra space beforehand. Thanks!

src/vergil/components/TaskList.java
riyadh-h: True, as it would improve the readability of the code.

src/vergil/components/Ui.java
riyadh-h: Right, I'll work on it.

src/vergil/main/Vergil.java
riyadh-h: I think it would be more suitable to split them into multiple sub-packages, as it would make the process of locating classes with strong associations easier.

I do agree, though, that some rearrangement is needed.

src/vergil/main/Vergil.java
riyadh-h: That would be true for some of the fields, but others, such as `storage` in the case of loading a different save file, would seem more convenient as variables.

Nonetheless, as far as our iP is concerned, there shouldn't be any harm in declaring them as `final`. Thanks for the tip!

src/vergil/testing/VergilTest.java
riyadh-h: Yes. Although there is only one test class at the moment, I'll try to add more in the future 😅.

src/vergil/types/Command.java
riyadh-h: Right, I'll work on them.

riyadh-h (commented):
Thank you for your helpful feedback! I will work on my code with your comments in mind 👍.

src/vergil/main/Vergil.java
riyadh-h: Yes, it would be better to include it in the `switch` structure.

PR #173, opened 2021-01-26
[Shaelyn Lam] iP

shaelynl:


src/main/java/DukeLevel.java.rej
```diff
@@ -0,0 +1,151 @@
+diff a/src/main/java/DukeLevel.java b/src/main/java/DukeLevel.java	(rejected hunks)
+@@ -1,3 +1,4 @@
++import java.io.*;
```
IceBear789, line 3: Using the wildcard import is discouraged in this module. Maybe you can consider importing the relevant features individually?

src/main/java/MisterDuke.java
```diff
@@ -0,0 +1,687 @@
+import java.io.*;
```
IceBear789: I've noticed that this file is very long. You may want to consider putting different classes in separate files.

src/main/java/MisterDuke.java
```diff
@@ -0,0 +1,687 @@
...
+            String cmd = tempStr.substring(7);
+            if (tempStr.contains("[T]")) {
+                ToDo toDoTask = new ToDo(cmd);
+                if (!checkForTick.contains(" ")) {
+                    toDoTask.markAsDone();
+                }
+                tasksArrayList.add(toDoTask);
+            } else if (tempStr.contains("[D]")) {
+                String[] strArray = cmd.split("by:", 2);
+                String inst = strArray[0].substring(0, strArray[0].length() - 2);
```
IceBear789: The variable "inst" is not clear to me. Perhaps use a clearer variable name?

src/main/java/DukeLevel.java.rej
```diff
@@ -0,0 +1,151 @@
...
+ import java.util.ArrayList;
+ import java.util.Scanner;
+ 
+@@ -15,6 +16,18 @@
+ 
+ public class DukeLevel {
```
IceBear789, line 9: This method feels a bit long to me. Maybe you can consider using helper methods or dividing this method into several smaller methods?

src/main/java/DukeLevel.java.rej
```diff
@@ -0,0 +1,151 @@
...
+         }
++
++        try {
++            saveData(tasksArray, dukeFile);
++        } catch (Exception e) {
++            e.getStackTrace();
++        }
++    }
++
++    private static void loadData(ArrayList<Task> tasksArrayList, File fname) throws IOException {
```
IceBear789, line 102: I think the variable "fileName" may be clearer here than "fname".

IceBear789 (commented):
You may want to consider putting different classes in different files so that the code is not too long. Other than that, the code is generally clear and easy to understand.

src/main/java/MisterDuke.java
```diff
@@ -0,0 +1,678 @@
...
+     */
+    public void run() throws IOException {
+        storage.load();
+        ui.printLine();
+        ui.showWelcome();
+        ui.printLine();
+        boolean isRunning = true;
+        while (isRunning) {
+            try {
+                String command = ui.readCommand();
```
chwoozy: Perhaps a better way to name it would be "textInput"? Since there is a "cmd" or Command type that could be misleading.

src/main/java/MisterDuke.java
```diff
@@ -0,0 +1,678 @@
...
+     * Loads list of tasks from the hard disk into the list of tasks
+     * @return array list of tasks
+     * @throws FileNotFoundException when the file cannot be found on the hard disk
+     */
+    public ArrayList<Task> load() throws FileNotFoundException {
+        Scanner sc = new Scanner(this.txtFile);
+        ArrayList<Task> tasksArrayList = new ArrayList<>();
+
+        while (sc.hasNextLine()) {
+            String tempStr = sc.nextLine();
```
chwoozy: I believe it would be better to name your variable fully ("tempString" instead of "tempStr") to improve readability!

src/main/java/MisterDuke.java
```diff
@@ -0,0 +1,678 @@
...
+     * @param command user input
+     * @return class that corresponds to the command
+     * @throws DukeException when the user input is wrong/incomplete
+     */
+    public static Command parse(String command) throws DukeException {
+        String input = command.trim();
+        String[] strArray = input.split(" ", 2);
+        String cmd = strArray[0].trim();
+
+        if (cmd.equalsIgnoreCase("todo")) {
```
chwoozy: Seeing that there are multiple if else cases, perhaps you may want to consider using a switch case? That would improve future developers to be able to figure out components easily by looking at the case conditions! 😃

src/main/java/MisterDuke.java
```diff
@@ -0,0 +1,678 @@
...
+class DukeException extends Exception {
+    public DukeException(String message) {
+        super(message);
+    }
+
+    @Override
+    public String toString() {
+        return super.getMessage();
+    }
+}
```
chwoozy, line 51: Perhaps you may want to separate the classes into different Java files? That would certainly declutter the amount of code in single Java file and makes it easier for developers to look at a specific area of code. Doing this would cause future extensions to increase the number of lines in a single class even more.

chwoozy (commented):
Code looks generally fine. Can possibily make it more readible with the choice of variable names.  Would also be great if different classes can be in different files. LGTM.

src/main/java/MisterDuke.java
```diff
@@ -0,0 +1,678 @@
+import java.io.*;
```
NightRaven49: Agreed with IceBear. Would also make program easier to edit rather than having to Crtl-F the class every time.

Also, can consider explicitly listing which packages are being imported rather than using wildcards.

src/main/java/MisterDuke.java
```diff
@@ -0,0 +1,678 @@
...
+        storage.save(tasks);
+    }
+
+    /**
+     * Main driver function
+     * @param args command line args
+     * @throws IOException when the input is wrong/incomplete
+     */
+    public static void main(String[] args) throws IOException {
+        new MisterDuke("src\\main\\data\\duke.txt").run();
```
NightRaven49: Not sure if this filepath will work on non Windows machines.

src/main/java/MisterDuke.java
```diff
@@ -0,0 +1,678 @@
...
+        try {
+            tasks = new ArrayList<>(storage.load());
+        } catch (IOException e) {
+            tasks = new ArrayList<>();
+        }
+    }
+
+    /**
+     * Runs Mister Duke
+     * @throws IOException when the input is wrong/incomplete
```
NightRaven49: Should there be a line break between description and parameters section? Can also look at other Javadocs.

src/main/java/MisterDuke.java
```diff
@@ -0,0 +1,678 @@
...
+}
+
+/**
+ * The Parser class takes in the user input, parses it,
+ * then display the corresponding message from Mister Duke
+ */
+class Parser {
+    private final Ui ui;
+    private Storage storage;
+    private ArrayList<Task> taskList;
```
NightRaven49: Should the name of the collection be in plural form instead?

NightRaven49 (commented):
Overall, did not find much coding errors, though due to the long nature of the file it became quite hard to keep focus. 🥺✌

PR #174, opened 2021-01-26
[Wei Yutong] iP

wei-yutong:


src/main/java/Duckie.java
```diff
@@ -0,0 +1,47 @@
+package main.java;
+
```
vevek: Should this extra line be removed?

src/main/java/Duckie.java
```diff
@@ -0,0 +1,47 @@
...
+
+import main.java.classes.*;
```
vevek: Should the * here be replaced with explicit import statements? I believe it's part of the coding standard.

src/main/java/Duckie.java
```diff
@@ -0,0 +1,47 @@
...
+import main.java.command.Command;
+
+import java.io.IOException;
+
+public class Duckie {
+    private Storage storage;
+    private TaskList lst;
```
vevek: Would it be better to rename "lst" to "list" here? Might be more understandable in the areas where it is used.

src/main/java/classes/Keyword.java
```diff
@@ -0,0 +1,11 @@
+package main.java.classes;
+
+public enum Keyword {
+    todo,
+    list,
+    deadline,
+    event,
+    done,
+    delete,
+    bye;
```
vevek: Since these enums are constants, should they be in upper case?

src/main/java/classes/Storage.java
```diff
@@ -0,0 +1,99 @@
...
+import java.io.IOException;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.Locale;
+import java.util.Scanner;
+
+public class Storage {
+    private File f;
```
vevek: Would it be better to rename "f" as "file" to make it more readable?

src/main/java/classes/Storage.java
```diff
@@ -0,0 +1,99 @@
...
+            } else {
+                fw.write(String.format("T | %s | %s",
+                        temp.getIsDone(), temp.getDescription()) + "\n");
+            }
+        }
+        fw.close();
+    }
+
+    public ArrayList<Task> loadTasks() throws IOException {
+        ArrayList<Task> AL = new ArrayList<>();
```
vevek: Perhaps it would be better to rename AL to arrayList or taskList here.

vevek (commented):
Hi Yutong! Overall I found your code easy to follow and read. In some places however I found the naming convention to deviate from the coding standards which made some parts difficult to understand. Consider making your enums uppercase as they are constants. Also some imports used *, where it might be better to list the imports out.

src/main/java/classes/Parser.java
```diff
@@ -0,0 +1,39 @@
...
+            case "deadline":
+                return new DeadlineCmd(cmd);
+            case "done":
+                return new DoneCmd(cmd);
+            case "delete":
+                return new DeleteCmd(cmd);
+            case "bye":
+                return new ByeCmd(cmd);
+            default:
+                throw new DuckieException("please start with 'list', 'todo', 'event', deadline', 'delete', 'done' or 'bye'");
```
tanboonji: Perhaps you can use your Keyword enum here since the cases are your Keyword enum values?

src/main/java/classes/Storage.java
```diff
@@ -0,0 +1,99 @@
...
+                    e.printStackTrace();
+                }
+
+                Event tempE = new Event(sArr[2].trim(), date);
+                if (sArr[1].trim().equals("done")) {
+                    tempE.isDone = true;
+                } else if (sArr[1].trim().equals("not done")) {
+                    tempE.isDone = false;
+                }
+                AL.add(tempE);
```
tanboonji: Perhaps the code inside the two if statements can be abstracted into another method? The code is almost the same in both if statements, with the exception that one is for creation of Deadline while the other is for creation of Event.

src/main/java/command/DeadlineCmd.java
```diff
@@ -0,0 +1,58 @@
...
+        //System.out.println(line);
+        String[] strD = cmd.trim().split("/by", 2);
+        String inputDate = strD[1].trim();
+        SimpleDateFormat format = new SimpleDateFormat("dd-MM-yyyy HHmm", Locale.ENGLISH);
+        Date date;
+        try {
+            date = format.parse(inputDate);
+        } catch (Exception e) {
+            throw new DuckieException("please enter date in the format dd-mm-yyyy HHmm");
+        }
```
tanboonji: This exact code which parses a String to Date object can be seen multiple times throughout your code. 

Perhaps would it be better if the method is abstracted into another class (eg. `DateParser`) and made into a static method? 

This way, you can still access the same code from your other classes and at the same time, improve the maintainability of your code as you would only be required to edit it once if it is abstracted into a separate method and class.

src/main/java/command/Event.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.util.Date;
+
+/**
+ * Event inherits from Task. It encapsulates an event with an additional String at.
+ * Event is specified by [E].
+ */
+public class Event extends Task {
+    protected Date at;
+    public Event (String description, Date at) {
```
tanboonji: There is a missing line break between the variable declaration and constructor. You might have missed it out by accident.

tanboonji (commented):
LGTM. Overall, I found your code easy to read for the most part except for a few places where the code gets repetitive. I suggested a few places where abstraction can be done to increase the maintainability and readability of your code.

src/main/java/classes/Storage.java
```diff
@@ -0,0 +1,117 @@
...
+            System.out.println("file loaded!");
+        }
+    }
+
+    /**
+     * Method to save Task in saved file, converts Task objects to Strings in file
+     * @param AL ArrayList of Task objects from user input
+     * @throws IOException if user IO is incorrect
+     */
+    public void saveTasks(ArrayList<Task> AL) throws IOException {
```
w-yuchen: Would AL be a bit confusing here? Perhaps `tasks` or `taskList` would be better?

src/main/java/classes/Parser.java
```diff
@@ -0,0 +1,73 @@
...
+                // Fallthrough
+            case "delete":
+                return new DeleteCmd(cmd);
+                // Fallthrough
+            case "find":
+                return new FindCmd(cmd);
+                // Fallthrough
+            case "bye":
+                return new ByeCmd(cmd);
+                // Fallthrough
```
w-yuchen: Since return also breaks out of the switch block and there are no fallthroughs here, would it be better if the `// Fallthrough` comments are removed?

src/main/java/classes/Storage.java
```diff
@@ -0,0 +1,117 @@
...
+        fw.close();
+    }
+
+    /**
+     * Method to load Tasks from saved file, converts Strings in saved file to Task objects.
+     * @return an ArrayList AL of Task objects
+     * @throws IOException if user IO is incorrect
+     */
+    public ArrayList<Task> loadTasks() throws IOException {
+        ArrayList<Task> AL = new ArrayList<>();
```
w-yuchen: I agree with vevek, arrayList or taskList could be better.

src/main/java/classes/Task.java
```diff
@@ -0,0 +1,55 @@
...
+     */
+    public String getIsDone() {
+        if (this.isDone) {
+            return "done";
+        } else {
+            return "not done";
+        }
+    }
+
+    public String getDescription() {
```
w-yuchen, line 40: You may have missed out the Javadoc comments for `getDescription()` here.

src/main/java/classes/Ui.java
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Method to read the user's input.
+     * @return String user input.
+     */
+    public String readInput() {
+        Scanner sc = new Scanner(System.in);
+        return sc.nextLine();
+    }
+
+    public void findMessage() {
```
w-yuchen: No Javadoc for `findMessage()`.

src/main/java/command/FindCmd.java
```diff
@@ -0,0 +1,33 @@
+package main.java.command;
+
+import main.java.classes.*;
+
+public class FindCmd extends Command {
+    private String cmd;
+
+    public FindCmd(String cmd) {
```
w-yuchen, line 13: Public constructor here with no Javadoc.

w-yuchen (commented):
Hi Yutong! Looks really great and easy to read! Just some nitpicks here and there.

PR #176, opened 2021-01-26
[Shaun Loh] iP

Juzzanoob:


src/main/java/FastIO.java
```diff
@@ -0,0 +1,82 @@
+import java.io.*;
```
NBH99: Should list out all the import class instead of using wildcard.

NBH99 (commented):
Good job in following the coding standard.

src/main/java/FastIO.java
Juzzanoob: Thank you for your input! I will make the changes accordingly.

src/main/java/Duke.java
```diff
@@ -1,10 +1,147 @@
...
+                }
+            }
+
+        } catch (DukeException e) {
+            System.out.println(e.errorMessage());
+
+        } finally {
+            reader.close();
+        }
+    }
```
swayongshen: Could this method be separated into multiple shorter methods?
https://nus-cs2103-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-long-methods

src/main/java/Event.java
```diff
@@ -0,0 +1,69 @@
...
+        if (this.isDone()) {
+            return "E 1 "
+                    + this.getTask_details()
+                    + " (on: " + date.format(DateTimeFormatter.ofPattern("d MMM yyyy")) + " "
+                    + time.format(DateTimeFormatter.ofPattern("HHmm")) + " )";
+        } else {
+            return "E 0 "
+                    + this.getTask_details()
+                    + " (on: " + date.format(DateTimeFormatter.ofPattern("d MMM yyyy")) + " "
+                    + time.format(DateTimeFormatter.ofPattern("HHmm")) + " )";
```
swayongshen: Would it be possible to simplify these expressions? 
https://nus-cs2103-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-complicated-expressions

src/main/java/Storage.java
```diff
@@ -0,0 +1,137 @@
...
+
+                output.add(toAdd);
+            }
+
+        } catch (FileNotFoundException e) {
+            throw new DukeException("loading error");
+        }
+
+        return output;
+    }
```
swayongshen: Would i be possible to shorten this method such that it is shorter than 1 page?

swayongshen (commented):
Great job on the clean code, just some minor reviewing that you might want to consider.

PR #178, opened 2021-01-26
[Toh Sihui] iP

tsh22:


src/main/java/Duke.java
```diff
@@ -1,10 +1,88 @@
...
+                    System.out.println("Did you miscount your tasks? "
+                            + "This task number isn't valid. Add it into my list!");
+                    continue;
+                }
+                //If valid, change status of task and output done task
+                Task item = currList.get(Integer.parseInt(commandArr[1]) - 1);
+                item.markAsDone();
+                System.out.println("Nice job! I've marked this task as done:\n"
+                        + item.toString());
+            } else if ((commandArr[0]).equals("delete")){
```
jlxw48: Do you think it will be better if you insert a whitespace between ')' and '{'?

src/main/java/Duke.java
```diff
@@ -1,10 +1,88 @@
...
+                Task item = currList.remove(Integer.parseInt(commandArr[1]) - 1);
+                System.out.println("Gotcha. I've removed this task:\n"
+                        + item.toString()
+                        + "\nNow you have " + currList.size() + " task(s) in the list.");
+            } else {
+                String taskType = commandArr[0];
+                String wholeTask;
+                Task item;
+                switch (taskType) {
+                    case "todo":
```
jlxw48: Do you think it will be more organised if the case blocks are indented at the same level as the switch block?

src/main/java/Duke.java
```diff
@@ -1,10 +1,88 @@
+import java.util.*;
```
jlxw48: Do you think it would be better to import only the necessary libraries?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
...
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = by;
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + "(" + by + ")";
+    }
+}
```
jlxw48: Remember to include newlines at the end of each file!

jlxw48 (commented):
Generally looks good! 
Minimal coding style errors.

src/main/java/Duke.java
```diff
@@ -35,9 +35,14 @@ public static void main(String[] args) {
                 item.markAsDone();
                 System.out.println("Nice job! I've marked this task as done:\n"
                         + item.toString());
+            } else if ((commandArr[0]).equals("delete")){
+                Task item = currList.remove(Integer.parseInt(commandArr[1]) - 1);
+                System.out.println("Gotcha. I've removed this task:\n"
+                        + item.toString()
+                        + "\nNow you have " + currList.size() + " task(s) in the list.");
```
jared98lyj: Shift the lines to make it top down (42 to 41)?

jared98lyj:
Generally good code structure. Good use of inheritance for the task class. Code is also neat without unecessary repeats. Some general formatting issues, but are minor.

PR #179, opened 2021-01-26
[Huang Nanxi] iP

Nanxi-Huang:


src/main/java/Parser.java
```diff
@@ -0,0 +1,95 @@
...
+                        "\n___________________________________________________________________\n"
+                                + "\u2639 OOPS!!! you need to enter the index of the task :-( \n"
+                                + "___________________________________________________________________\n");
+            }
+
+        } else if (description.toLowerCase().contains(deleteCommand)) {
+            try {
+                int taskIndex = Integer.parseInt(
+                        description.replaceAll("[^0-9]", ""));
+                tasks.Delete(taskIndex);
```
zatkiller: Should Delete method begin with small capital letter?

src/main/java/TaskList.java
```diff
@@ -0,0 +1,225 @@
...
+                    + "Got it. I've added this task: \n"
+                    + myTask.toString());
+            Remark();
+        }
+    }
+
+    /**
+     * Prints the number of tasks in the list.
+     */
+    public void Remark() {
```
zatkiller: should Remark begin with a small capital letter as well?

zatkiller (commented):
Overall code is easy to read with no major code style violation. Good work!

mechastriker3:
Great Job Nanxi!! The code style looks ok in general :)

src/main/java/Duke.java
```diff
@@ -17,16 +17,6 @@ public Duke(String filePath) {
     }
 
 
-
-    public void Greet() {
```
mechastriker3: Should the method name start with small letters? So would it be better if it was `greet()` instead of `Greet()`

src/main/java/Duke.java
```diff
@@ -35,17 +25,6 @@ public void run() throws InvalidCommandException {
     }
 
 
-    public void Exit() {
```
mechastriker3: Same for this method, should the method name be `exit()` instead of `Exit()`

mechastriker3 (commented):
Hi Nanxi, great job!! Generally the code is well written with no major code style violations.

src/main/java/Duke.java
Nanxi-Huang: Thanks for the comment. Will change the method name to start with small letters!

src/main/java/Duke.java
Nanxi-Huang: Thanks for pointing out!

PR #180, opened 2021-01-26
[Liao Xing Peng] iP

mechastriker3:
Testing Pull Request


src/main/java/Duke.java
```diff
@@ -1,10 +1,25 @@
+import java.io.IOException;
+import java.util.*;
+import duke.*;
```
zatkiller: Should we avoid using wildcard import?

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,27 @@
...
+        this.endTime = endTime;
+    }
+
+    public Deadline(String taskDescription, LocalDate endTime, boolean isDone) {
+        super(taskDescription, isDone);
+        this.endTime = endTime;
+    }
+
+    public String getTaskType() {
+
```
zatkiller: Should we remove the blank lines here?

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,28 @@
+package duke;
+
+import java.time.LocalDate;
+
+public class Event extends Task{
+    private static String taskType = "EVENT";
+//    private String startTime;
```
zatkiller: Should we remove these comments since it's not used as an attribute of the class or for documentation purposes?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+
+    public void writeData(List<Task> taskList) throws IOException {
+        FileWriter file = new FileWriter(DEFAULT_PATH);
+        for (Task task : taskList) {
+            String toBeSaved = "";
+            switch (task.getTaskType()) {
+                case "DEADLINE":
```
zatkiller: Is the switch statement case indentation supposed be as such?

zatkiller (commented):
Overall, code is easy to understand except for some code styling issues that can be fixed. Good work!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,137 @@
...
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+        }
+    }
+
+    /**
+     * Method to save TaskList to hard disk.
+     *
+     * @param taskList
```
justgnohUG: Perhaps consider standardizing on having or omitting "\n" between JavaDoc description and @param etc.

src/main/java/duke/Storage.java
justgnohUG: Second this. I believe your switch-case statements in your Parser were indented as per Java coding standards.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,43 @@
...
+        this.isDone = done;
+    }
+
+    public boolean getIsDone() {
+        return isDone;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "\u2713" : "\u2718"); //return tick or X symbols
+    }
```
justgnohUG, line 45: Perhaps the comment could be written as a JavaDoc instead!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,137 @@
...
+import java.io.FileWriter;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Storage class to save information to hard disk.
+ */
```
justgnohUG, line 22: Consider utilising other tags like @author

src/main/java/Duke.java
```diff
@@ -1,10 +1,49 @@
...
+import duke.Parser;
+import duke.Storage;
+import duke.TaskList;
+import duke.Ui;
+import duke.command.Command;
+import duke.exception.DukeCommandException;
+import duke.exception.DukeException;
+import duke.exception.DukeToDoException;
+
 public class Duke {
```
justgnohUG, line 11: Overall, the code quality was great!, besides a couple of minute details, and I discovered different encapsulations and OOP use cases from your project. Keep up the good work!

src/main/java/Duke.java
mechastriker3: Thanks for catching this!!

src/main/java/duke/Deadline.java
mechastriker3: Agreed!!

src/main/java/duke/Storage.java
mechastriker3: Yup thanks for catching this!! :)

src/main/java/duke/Storage.java
mechastriker3: Good Point! Thanks for the tip :)

PR #186, opened 2021-01-26
[Lee Rong Jieh] iP

lrj689:


src/main/java/Deadlines.java
```diff
@@ -0,0 +1,14 @@
+public class Deadlines extends Task {
+
+    private String by;
```
banchiang: Similiar mistake as me! We can change the variable name 'by' to be more intuitive

src/main/java/Duke.java
```diff
@@ -1,10 +1,115 @@
...
-        System.out.println("Hello from\n" + logo);
+        String bound = "____________________________________________________________";
+        String topBound = "____________________________________________________________\n";
+        String bottomBound = "\n____________________________________________________________\n";
+
+        System.out.println(topBound + logo + "\nHello! I'm Duke\n" + "What can I do for you?\n" + bottomBound);
+
+        while (sc.hasNextLine()) {
+            String input = sc.nextLine();
+            String[] arr = input.split("/");
```
banchiang: String array name can be more descriptive :D

src/main/java/Duke.java
```diff
@@ -1,10 +1,115 @@
...
+                    System.out.println(topBound + "Task does not exist and cannot be deleted!" + bottomBound);
+                } catch (NumberFormatException e) {
+                    // parameter for delete is not an Integer
+                    System.out.println(topBound + "delete must be followed by an Integer, ie. delete 1. Try again!"
+                        + bottomBound);
+                }
+            } else {
+                // unknown command
+                System.out.println(topBound + "OOPS!!! I'm sorry, but I don't know what that means :-(" + bottomBound);
+            }
```
banchiang: Since there are many if else statements, I think using switch would make the code look cleaner :)

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
+public class Task {
+    /**
+     * Task class stores a String as name of the task and boolean indicating task completion
+     */
+
+    private String taskName;
+    private boolean done;
```
banchiang: As per coding standard, boolean variable name would be a name to point out that the variable is boolean. E.g ("isDone") would be good :D

banchiang (commented):
Review for ip for coding standards. Do have a look :D

src/main/java/Deadlines.java
SoulUseless: Since this class indicates a singular Deadline Task object, it could be better if you named your class in the singular form instead of the plural form. This also reduces confusion should the need for having a Deadline array arise.

src/main/java/Duke.java
```diff
@@ -1,10 +1,115 @@
...
+                    System.out.println(topBound + "Good job! I've marked this task as done:\n" + "  " + list.get(x)
+                            + bottomBound);
+                } catch (IndexOutOfBoundsException e) {
+                    // Task number x does not exist
+                    System.out.println(topBound + "Task does not exist and can not be completed!" + bottomBound);
+                } catch (NumberFormatException e) {
+                    // parameter for done is not an Integer
+                    System.out.println(topBound + "done must be followed by an Integer ie. done 1. Try again!"
+                        + bottomBound);
+                }
```
SoulUseless: I have noticed that you have implemented DukeException.java as a error handling class. Perhaps you can leverage that to handle your exceptions instead of printing them here?

src/main/java/Duke.java
```diff
@@ -1,10 +1,115 @@
...
+                    System.out.println(topBound + "delete must be followed by an Integer, ie. delete 1. Try again!"
+                        + bottomBound);
+                }
+            } else {
+                // unknown command
+                System.out.println(topBound + "OOPS!!! I'm sorry, but I don't know what that means :-(" + bottomBound);
+            }
+        }
     }
 }
```
SoulUseless: I like how thorough your error handling is. However, the main function is very long. Would it be better to abstract some if-block code into separate functions to aid readability?

src/main/java/DukeList.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public DukeList() {
+        this.list = new ArrayList<Task>();
+    }
+
+    public void add(Task item) {
+        list.add(item);
+    }
+
+    public void done(int x) {
```
SoulUseless: Since this is a function that does an action, would it be better to name it in a way that it sounds like a function?

src/main/java/Events.java
```diff
@@ -0,0 +1,15 @@
...
+    private String duration;
+
+    public Events(String name, String duration) {
+        super(name);
+        this.duration = duration;
+    }
+
+    @Override
+    public String toString() {
+        return "[E]" + super.toString() + " (at: " + duration + ")";
```
SoulUseless: I like your use of the inherited toString() method.

SoulUseless (commented):
Please find below my PR review.

src/main/java/Duke.java
```diff
@@ -1,10 +1,190 @@
...
+
 public class Duke {
     public static void main(String[] args) {
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        String bound = "____________________________________________________________";
```
YuFeng0930: Constant names must be all uppercase.

src/main/java/Duke.java
```diff
@@ -1,10 +1,190 @@
...
+                    System.out.println(topBound
+                            + "Uh Oh! show must be followed by a date in YYYY-MM-DD format.\nTry again!" + bottomBound);
+                }
+            } else {
+                // unknown command
+                System.out.println(topBound + "Uh Oh! I'm sorry, but I don't know what that means." + bottomBound);
+            }
+        }
+    }
+    public static void readFromFile(String filePath, DukeList list) throws FileNotFoundException {
```
YuFeng0930: There should be a line break to separate functions. JavaDoc for public functions need to be added also.

src/main/java/Duke.java
```diff
@@ -1,10 +1,190 @@
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.util.Scanner;
+
 public class Duke {
     public static void main(String[] args) {
```
YuFeng0930: Cannot write everything in main function. Even if it's not main function, this function itself is too long also, which reduce the readability of the code.

src/main/java/DukeList.java
```diff
@@ -0,0 +1,103 @@
...
+    public int noOfTasks() {
+        return list.size();
+    }
+
+    /**
+     *
+     * @param x Task number to be removed
+     * deletes the stated task
+     */
+
```
YuFeng0930: There shouldn't be a line break between JavaDoc and function header.

src/main/java/Duke.java
YuFeng0930: Agree with you, since the scope of arr is quite big.

YuFeng0930 (commented):
The first thing needs to be done is move things out of main function. Try to break each functions into more functions, which will make the code much understandable and easier to maintain.

PR #187, opened 2021-01-26
[Winnie Ho Yi Xuan] iP

Winniehyx:


src/main/java/Duke.java
```diff
@@ -1,10 +1,294 @@
+import task.*;
```
vevek: Should the * import can be replaced with the individual required imports? I think it was in the coding standard guidelines.

src/main/java/Duke.java
```diff
@@ -1,10 +1,294 @@
...
+import storage.Storage;
+import ui.Ui;
+
+import java.util.*;
```
vevek: Should the * import can be replaced with the individual required imports? I think it was in the coding standard guidelines.

src/main/java/Duke.java
```diff
@@ -1,10 +1,294 @@
...
+    private Storage storage;
+    private TaskList taskList;
+    private Ui ui;
+
+    public Duke() {
+
+        this.ui = new Ui();
+        this.storage = new Storage(System.getProperty("user.dir") + "/data/", "duke.txt");
+        this.taskList = storage.load();
+
```
vevek: Should the extra lines between the statements and the method braces exist? It might make the code a little harder to read.

src/main/java/Duke.java
```diff
@@ -1,10 +1,294 @@
...
+    }
+
+    public void run() {
+
+        Scanner sc = new Scanner(System.in);
```
vevek: As above, perhaps some of the extra lines between statements can be removed and  the statements can be grouped instead. What do you think?

src/main/java/Duke.java
```diff
@@ -1,10 +1,294 @@
...
+
+        ui.printWelcomeMessage();
+
+        boolean isExit = false;
```
vevek: Would it be better to rename isExit to hasExited or something else? I'm not too sure what it refers to.

src/main/java/Duke.java
```diff
@@ -1,10 +1,294 @@
...
+        while(sc.hasNextLine()) {
+
+            String input = sc.nextLine();
+
+            String[] tokens = input.split(" ", 2);
+
+            String command = tokens[0];
+            String argument = tokens.length == 2 ? tokens[1] : null;
+
+            if (command.equals("bye")) {
```
vevek: Perhaps the code here can be replaced with a switch statement to improve readability. What do you think?

src/main/java/Duke.java
```diff
@@ -1,10 +1,294 @@
...
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+
+        new Duke().run();
+
```
vevek, line 81: Should there be an extra line here? Noticed it in a few places.

src/main/java/Duke.java
```diff
@@ -1,10 +1,294 @@
...
+    /**
+     * Executes Done Command
+     *
+     * @param argument  command arguments
+     * @return          boolean to represent termination of the program
+     */
+    private boolean executeDoneCommand(String argument) {
+        // check for correct number of arguments
+        if (argument == null) {
```
vevek: I've noticed these lines being repeated in a few of the executeXXXXXCommand methods. Perhaps it would be better to abstract them out to make the overall code cleaner.

src/main/java/Duke.java
```diff
@@ -1,10 +1,294 @@
...
+            return false;
+        }
+
+        String[] split = argument.split("/at", 2);
+        String description = split[0];
+        String at = split.length > 1 ? split[1] : null;
+        Task task = new Events(description, at);
+        taskList.add(task);
+
+        System.out.println("Got it. I have added this task:");
```
vevek: Perhaps these bunch of lines can be abstracted out too as well. They appear repeated in the other executeXXXXCommand methods as mentioned above.

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,178 @@
...
+public class Storage {
+
+    private String directory;
+    private String fileName;
+
+    public Storage(String directory, String fileName) {
+        this.directory = directory;
+        this.fileName = fileName;
+    }
+    /**
```
vevek, line 30: Should there be an extra line before this line?

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,178 @@
...
+     * Saves the tasklist.
+     * Converts a tasklist into its string representation then saves it in a file
+     *
+     * @param taskList  the tasklist to be saved
+     */
+    public void save(TaskList taskList) {
```
vevek: Perhaps the method name "save" could be renamed to "saveTaskList" or "saveToHarddisk".

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,178 @@
...
+            e.printStackTrace();
+        }
+
+    }
+    /**
+     * Loads a text file and parses it into a tasklist
+     *
+     * @return
+     */
+    public TaskList load() {
```
vevek: Similar to the previous comment, perhaps "load" can be renamed to "loadFromHarddisk" or "loadTasklist".

vevek (commented):
Hi Winnie! Overall, I found the code easy to follow and read. There was some minor coding standard violations and perhaps some areas where the code quality could be improved. Some methods had repeated code which can be abstracted out to make the overall code simpler to read too.

src/main/java/Duke.java
Winniehyx: Got it! Thank you for pointing it out :)

src/main/java/Duke.java
Winniehyx: Yup! thank you

src/main/java/Duke.java
Winniehyx: In my opinion, its needed because I am initializing new objects.

src/main/java/Duke.java
Winniehyx: Changed :)

src/main/java/Duke.java
Winniehyx: I wanted to have a boolean instead of a loop but I have already removed this in my recent code.

src/main/java/Duke.java
Winniehyx: Thank you for the idea! I will try to implement it

src/main/java/Duke.java
Winniehyx: Thank you for pointing it out!

src/main/java/storage/Storage.java
Winniehyx: Thank you for pointing it out!

src/main/java/Duke.java
Winniehyx: Thank you for the suggestion!

src/main/java/Duke.java
Winniehyx: Thank you for the suggestion!

src/main/java/storage/Storage.java
Winniehyx: Changed to what you have suggested! Thank you!

src/main/java/storage/Storage.java
Winniehyx: Amended to what you have suggested

PR #189, opened 2021-01-26
[Douglas Wei Jing Allwood] iP

douglaswja:


src/main/java/dbot/command/Command.java
```diff
@@ -0,0 +1,150 @@
...
+    public Command(int targetIndex) {
+        setTargetIndex(targetIndex);
+    }
+
+    /**
+     * Sets the target index instance variable.
+     *
+     * @param targetIndex A integer specifying the new target index for the 0-indexed TaskList.
+     */
+    protected void setTargetIndex(int targetIndex) {
```
CSjiade: Hi sir for this function should it be public too just like the rest of the functions?

PS: I may be wrong here

src/main/java/dbot/command/Command.java
```diff
@@ -0,0 +1,150 @@
...
+        this.targetIndex = targetIndex;
+    }
+
+    /**
+     * Returns the target index for the 0-indexed TaskList.
+     *
+     * @return An integer specifying the target index for the 0-indexed TaskList.
+     */
+    protected int getTargetIndex() {
```
CSjiade: Hi sir for this function should it be public too just like the rest of the functions?

PS: I may be wrong here

src/main/java/dbot/storage/Storage.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+    }
+
+    /**
+     * Saves the given TaskList by writing it to a file, creating one if one does not already exist.
+     *
+     * @param tasks The TaskList to be saved.
+     * @throws StorageOperationException If the file cannot be written to.
+     */
+    public void save(TaskList tasks) throws StorageOperationException {
```
CSjiade: Will saveTasks be better here?

CSjiade (commented):
Overall the code is very neat and good use of classes and oop. I think it's all good! Dont really have to edit anything. Good job! :)

src/main/java/dbot/command/Command.java
```diff
@@ -0,0 +1,150 @@
...
+     * @return A String containing the description that was supplied to this Command.
+     */
+    public String getDescription() {
+        return description;
+    }
+
+    /**
+     * Returns true if this Command is an Exit Command and false otherwise.
+     *
+     * @return True if this is an Exit Command and false otherwise.
```
spzj: Hi Douglas! You might want to edit the javadocs for this since it only returns false.

src/main/java/dbot/command/FindCommand.java
```diff
@@ -0,0 +1,37 @@
...
+        for (Task task : tasks) {
+            if (task.getDescription().contains(keyword)) {
+                relevantTasks.add(task);
+            }
+        }
+        ui.printRelevantTasks(relevantTasks);
+    }
+
+    @Override
+    public void quietExecute(TaskList tasks, Storage storage) throws DBotException {
```
spzj: Since it's not a DBotException, you can remove "throws DBotException".

spzj (commented):
Generally, I find your code really neat and well-documented. Great job!

src/main/java/dbot/command/Command.java
douglaswja: Hi, thank you for the review.
You are perhaps right that the Setter can be public, ill look into changing it.

src/main/java/dbot/command/Command.java
douglaswja: Yes, ill look into changing this as well, thank you.

src/main/java/dbot/storage/Storage.java
douglaswja: Yes, that would be a more expressive name for the method and ill amend it accordingly.

src/main/java/dbot/command/Command.java
douglaswja: Hi, thank you for the review.
This method was defined to return false in the Command class so that all classes that inherit it would also return false unless they override it, which is done only in the ExitCommand class.

I do understand the confusion that this method can cause and recognise that a more elegant solution might be to use a static method in the ExitCommand class to check if the user input is equivalent to an exit command.

(An example implementation that gave me the above idea can be found at: https://github.com/se-edu/addressbook-level2/blob/master/src/seedu/addressbook/commands/ExitCommand.java)

src/main/java/dbot/command/FindCommand.java
douglaswja: Thank you for pointing this out! Might have missed this, will look into it.

PR #191, opened 2021-01-26
[Shen Yang] iP

goatygoatygoat:


src/main/java/Duke.java
```diff
@@ -1,10 +1,177 @@
...
+        duke.chat(scanner);
+    }
+
+    abstract class Task {
+        abstract Task done();
+    }
+
+    class ToDo extends Task {
+        private String name;
+        private boolean done;
```
IanCKW: Should this be isDone

src/main/java/Duke.java
```diff
@@ -1,10 +1,177 @@
...
+            this.done = false;
+        }
+
+        Deadline(String name, String deadline, boolean done) {
+            this.name = name;
+            this.deadline = deadline;
+            this.done = done;
+        }
+
+        @Override
```
IanCKW: Should be a verb right?

src/main/java/Duke.java
```diff
@@ -1,10 +1,177 @@
...
+                return String.format("[D][X] %s (by:%s)", this.name, this.deadline);
+            }
+            return String.format("[D][ ] %s (by:%s)", this.name, this.deadline);
+        }
+    }
+
+    class Event extends Task {
+        private String name;
+        private String duration;
+        private boolean done;
```
IanCKW: Should this be in abstract class?

src/main/java/Duke.java
```diff
@@ -1,10 +1,177 @@
...
+                    System.out.println(ice);
+                }
+            } else if (input.startsWith("event")) {
+                try {
+                    int index = input.indexOf("/at");
+                    Task newtask = new Event(input.substring(6, index - 1), input.substring(index + 3));
+                    tasks.add(newtask);
+                    System.out.println(String.format("Got it. I've added this task:\n%s\nNow you have %s tasks in the list.", newtask.toString(), tasks.size()));
+                } catch (Exception e) {
+                    IncompleteCommandException ice = new IncompleteCommandException();
```
IanCKW: should this be named differently?

IanCKW (commented):
There wasn't much to say since the code is still based on week 1's tasks. But overall, only a few minor issues!

src/main/java/ExitCommand.java
```diff
@@ -0,0 +1,38 @@
...
+     * @param obj the object which will be compared to this.
+     * @return Indication of whether obj is equivalent to this.
+     */
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == this) {
+            return true;
+        } else if (obj instanceof ExitCommand) {
+            return true;
+        }
```
lue97, line 27: perhaps you could just shorten this conditional to
`obj == this || obj instanceof ExitCommand`
?

lue97 (commented):
Very neat code (there wasn't much that issues that I could find).

PR #193, opened 2021-01-26
[Warren Ng] iP

NightRaven49:


src/main/java/Duke.java
```diff
@@ -1,10 +1,144 @@
...
+
+        userInput = new TextField();
+        sendButton = new Button("Send");
+
+        AnchorPane mainLayout = new AnchorPane();
+        mainLayout.getChildren().addAll(scrollPane, userInput, sendButton);
+
+        stage.setTitle("Duke");
+        stage.setResizable(false);
+        stage.setMinHeight(600.0);
```
stein414: Perhaps you could abstract the numbers out to a static final variable to avoid magic numbers?

src/main/java/Duke.java
```diff
@@ -1,10 +1,144 @@
...
+            dukeText = ui.bye();
+        } else {
+            dukeText = parser.parse(input);
+        }
+        dialogContainer.getChildren().addAll(
+                DialogBox.getUserDialog(userText, new ImageView(user)),
+                DialogBox.getDukeDialog(dukeText, new ImageView(duke))
+        );
+        userInput.clear();
+        if (input.equals("bye")) {
```
stein414, line 127: Good catch for the bye command to show the message before the program closes

src/main/java/Parser.java
```diff
@@ -0,0 +1,121 @@
...
+    public Parser(TaskList list, Ui ui) {
+        this.list = list;
+        this.ui = ui;
+    }
+
+    /**
+     * Main logic of the program.
+     * @return Label with response from program.
+     */
+    public Label parse(String input) {
```
stein414, line 27: Perhaps you should abstract the different commands to their respective functions to avoid the long `parse` method.

src/main/java/Parser.java
```diff
@@ -0,0 +1,121 @@
...
+     * @param ui Ui object for messages.
+     */
+    public Parser(TaskList list, Ui ui) {
+        this.list = list;
+        this.ui = ui;
+    }
+
+    /**
+     * Main logic of the program.
+     * @return Label with response from program.
```
stein414, line 25: You might want to add a blank line in between the description and @return. Reference for code style here [https://se-education.org/guides/conventions/java/index.html#comments](https://se-education.org/guides/conventions/java/index.html#comments)

Similarly for all other javadocs

src/main/java/Task.java
```diff
@@ -0,0 +1,103 @@
...
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/**
+ * Task object for CS2103T iP. Stores relevant data regarding a task, including type, task,
+ * date and time (if required) and whether it has been completed.
+ */
+public class Task {
+    private final String task;
+    private boolean done = false;
```
stein414: Maybe it would be better to call it ìsDone`for boolean variables?

src/main/java/Task.java
```diff
@@ -0,0 +1,103 @@
...
+     */
+    public void markDone() {
+        this.done = true;
+    }
+
+    /**
+     * Checks for whether the task is done. Used as part of the toString() method.
+     * @return "[ ]" for not done, "[X]" for done.
+     */
+    public String checkDone() {
```
stein414, line 55: Perhaps you should rename the method as the name does not reflect its behaviour?

src/main/java/Duke.java
```diff
@@ -1,10 +1,144 @@
...
+
+import javafx.scene.image.Image;
+import javafx.scene.image.ImageView;
+
+/**
+ * The Duke class is the main program responsible for instantiating all required classes
+ * and methods.
+ */
+public class Duke extends Application {
+    private final Storage ine;
```
xiinweii98, line 22: I think it will be good to give the Storage object a proper name.

src/main/java/Parser.java
xiinweii98: Else, you should probably use a switch case instead of many else if blocks. Remember to indent correctly if you do go with switch case.

src/main/java/Storage.java
```diff
@@ -0,0 +1,75 @@
...
+     */
+    public Storage(TaskList tasks, Ui ui) {
+        this.storage = tasks;
+        this.ui = ui;
+        data = new File("data/savedList.txt");
+        if (!data.exists()) {
+            try {
+                data.getParentFile().mkdirs();
+                data.createNewFile();
+                ui.print("It appears you are using Duke for the first time. Welcome!");
```
xiinweii98: Nice additional messages

src/main/java/Storage.java
```diff
@@ -0,0 +1,75 @@
...
+        sc.close();
+        return ui.print("Tasks saved from last session imported! :)");
+    }
+
+    /**
+     * Exports saved tasks.
+     * @throws IOException When there are IO errors.
+     */
+    public void exportData() throws IOException {
+        FileWriter fw = new FileWriter("data/savedList.txt");
```
xiinweii98, line 69: Maybe you could retrieve "data/savedList.txt" as a method argument instead of hard coding it right here. So, maybe use the method header instead:
public void exportData(String path) throws IOException

src/main/java/Task.java
```diff
@@ -0,0 +1,103 @@
...
+import java.time.format.DateTimeParseException;
+
+/**
+ * Task object for CS2103T iP. Stores relevant data regarding a task, including type, task,
+ * date and time (if required) and whether it has been completed.
+ */
+public class Task {
+    private final String task;
+    private boolean done = false;
+    private final int type;
```
xiinweii98, line 12: You can probably convert type into an enumeration if you want to.

xiinweii98 (commented):
Nice coding, was easy to read.

PR #196, opened 2021-01-26
[Princess Priscilla] iP

pPris:


src/main/java/Duke.java
```diff
@@ -1,10 +1,36 @@
...
+            Ui.print(new String[]{"Something went wrong in loading the task file and parsing",
+                    e.getMessage()});
+        }
+
+        Ui.intro();
+
+        // variables to reuse
+        String userInput;
+        Parser parser = new Parser(taskList);
+        boolean remainOpen = true;
```
geraldfan: Perhaps this variable name could sound more boolean?

src/main/java/Event.java
```diff
@@ -0,0 +1,58 @@
+import static java.lang.Boolean.parseBoolean;
+
+import java.time.LocalDateTime;
+
+public class Event extends Task {
```
geraldfan: Might be better to add a javadoc for this class?

src/main/java/ParseDateTime.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private static String unparseFormat1(LocalDateTime d) {
+        String patt = "dd-MM ha yyyy";
+        DateTimeFormatter f2 = DateTimeFormatter.ofPattern(patt);
+        String str = f2.format(d);
+        return str.substring(0, patt.length() - 4);
+    }
+
+    // where s is formatted like "30-04 6PM"
+    private static LocalDateTime parseFormat1(String s) {
```
geraldfan: Perhaps this method name could be more descriptive?

src/main/java/Storage.java
```diff
@@ -0,0 +1,119 @@
...
+                taskList.add(t);
+                isAnyTaskFound = true;
+            }
+        } else {
+            // probably not gonna be used due to TaskList.java impl
+            setupTasksFile();
+        }
+        return isAnyTaskFound;
+    }
+}
```
geraldfan, line 156: I like how the code has followed the java coding standard.

src/main/java/Todo.java
```diff
@@ -0,0 +1,42 @@
+import static java.lang.Boolean.parseBoolean;
+
+public class Todo extends Task {
```
geraldfan: Perhaps there could be a javadoc for this class?

geraldfan (commented):
Overall, I found the code to be easy to read. I noted few coding violations, but the code mostly followed the java coding standard and was well written.

src/main/java/Duke.java
pPris: Thanks for your review, all your comments were helpful. Will be adjusting my code accordingly soon.

src/main/java/Duke.java
```diff
@@ -1,10 +1,90 @@
...
+    // javafx code adapted/taken from https://se-education.org/guides/tutorials/javaFxPart3.html
+
+
+    /**
+     * This is the main function that generates a response to user inputted commands and strings.
+     *
+     * @param input User input
+     * @return Response after processing the command(s) and argument(s) from the user input
+     */
+    String getResponse(String input) {
```
rajobasu, line 50: Just wondering, any reason for this to be not public or private and default instead? Usually its preferred to mark methods as private/public unless specifically intended to work with similar classes in the same package.

src/main/java/Parser.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * Calls the appropriate command constructor based on the first word in the user input
+     * @param firstWord Corresponds to a command name or a short-form
+     * @param commandBody Arguments provided for the command
+     * @return A command that can be run
+     * @throws UnsupportedCommandException
+     */
+    private Command parseIntoCommand(String firstWord, String commandBody) throws UnsupportedCommandException {
```
rajobasu, line 47: An alternate way as to how I had implemented it was keep an Enum of Commands and then have a function called something like ```matchCommandString()``` so that the matching itself is also done with the class itself, makes it a bit more contained i guess? And then instead of a sswitch case statement we can just run a loop and match.

rajobasu (commented):
Hey just a few comments, thought might be a little different from your solving style ?

PR #197, opened 2021-01-26
[Jason Tay] iP

jay9645:


src/main/java/Seashell.java
```diff
@@ -0,0 +1,125 @@
...
+            throw new SeashellException("OOPS!!! Task list is currently still empty!");
+        } else {
+            for (int i = 1; i <= taskList.size(); i++) {
+                Task t = taskList.get(i - 1);
+                System.out.println(i + ". " + t);
+            }
+        }
+    }
+
+    private void done(String command, ArrayList<Task> taskList) throws SeashellException {
```
eksinyue: Maybe a better alternative for this method name is `MarkAsDone`? as names representing methods should be present tense verbs.

src/main/java/Seashell.java
```diff
@@ -0,0 +1,125 @@
...
+        String num = command.substring(5);
+        int taskNum = Integer.parseInt(num);
+        if (taskList.size() < taskNum) {
+            throw new SeashellException("OOPS!!! Task does not exist!");
+        } else {
+            Task updated = taskList.get(taskNum - 1).setDone();
```
eksinyue: Do you think a good alternative for this variable name would be `updatedTask`?

src/main/java/Task.java
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    public Task setDone() {
+        Task doneTask = new Task(this.name);
+        doneTask.isDone = true;
+        return doneTask;
+    }
+
+    public String getStatusIcon() {
+        return (this.isDone ? "\u2713" : "\u2718"); //return tick or X symbols
```
eksinyue: Just a suggestion, but maybe a good practice would be to make tick and X symbols constants. for e.g. `TICK_SYMBOL = "\u2713"` and `CROSS_SYMBOL = "\u2718"` to avoid having magic strings in the code.

src/main/java/Seashell.java
```diff
@@ -0,0 +1,125 @@
...
+            taskList.set(taskNum - 1, updated);
+            System.out.println("Nice! I've marked this task as done: \n" + updated);
+        }
+    }
+
+    private void delete(String command, ArrayList<Task> taskList) throws SeashellException {
+        String num = command.substring(7);
```
eksinyue: Instead of using `.substring()` with magic numbers, maybe you can consider using `string.split(" ", 2)[1]` to get the string after the first space.

eksinyue (commented):
Overall, I think you did a good job following the Java coding standard! 👍

rjeez (commented):
I think that the string name could be dueDate instead of by as it would be more understandable that way!
Capture

src/main/java/Seashell.java
Juzzanoob: Perhaps delete could be renamed to deleteTask too, just to ensure clarity?

Juzzanoob (commented):
Overall, I like the creative direction with naming the app as Seashell! :)

There is also much attention paid to ensuring the code follows the coding standard so well done!

PR #198, opened 2021-01-26
[ZHANG GUANGZESHENG] iP

Andrewzhang217:


src/main/java/Duke.java
```diff
@@ -1,10 +1,124 @@
...
+                        System.out.println(index + "." + deadline.toString());
+                    } else {
+                        System.out.println(index + "." + t.toString());
+                    }
+
+                }
+    }
+
+    public void markDone(int index){
+
```
jamesleeht: No empty line here necessary as well as some other areas.

src/main/java/Ui.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class Ui {
+
+    public static List<Task> list = new ArrayList<>();
+    public static void greeting() {
+        System.out.println("Hello! I'm Duke");
+        System.out.println("What can I do for you?");
+    }
+
+    public static void bye() {
```
jamesleeht: Maybe could rename to sayBye() instead to make it closer to a verb?

jamesleeht (commented):
Nicely done, naming is good! Pointed out a few small potential edits.

hojiefeng:
Overall, not a lot of coding quality issues.

src/main/java/Event.java
```diff
@@ -0,0 +1,14 @@
...
+    protected String by;
+
+    public Event(String description, String by) {
+        super(description);
+        this.by = by;
+    }
+
+    @Override
+    public String toString() {
+        return "[E]" + super.toString() + " (by: " + by + ")";
```
hojiefeng, line 12: Would a String.printf be cleaner here?

src/main/java/Task.java
```diff
@@ -32,4 +32,8 @@ public String getStatusIcon() {
     }
 
 
+    public String toString(){
+        return "[" + this.getStatusIcon() + "]" + this.getDescription();
```
rhgndf: Same comment as before.

src/main/java/Todo.java
```diff
@@ -0,0 +1,14 @@
...
+    protected String by;
+
+    public Todo(String description, String by) {
+        super(description);
+        this.by = by;
+    }
+
+    @Override
+    public String toString() {
+        return "[T]" + super.toString() + " (by: " + by + ")";
```
rhgndf: Also the same as before

rhgndf (commented):
Possibly cleaner code by using printf

src/main/java/FindCommand.java
```diff
@@ -0,0 +1,14 @@
+public class FindCommand extends Command {
+    private final String searchString;
+
+    private FindCommand(String searchString, Duke duke) {
+        super(-1, duke);
```
rhgndf: Magic number, could use an enum or static variable in the class

rhgndf (commented):
Magic number

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,51 @@
...
+        FileWriter fileWriter = new FileWriter(String.valueOf(this.filePath));
+        fileWriter.write(fileContent);
+        fileWriter.close();
+    }
+
+    private ArrayList<String> ReadFile(String fileName) throws FileNotFoundException {
+        Scanner scanner = new Scanner(fileName);
+        ArrayList<String> fileContent = new ArrayList<>();
+
+        while(scanner.hasNext()){
```
nighoggDatatype:
```suggestion
        while (scanner.hasNext()) {
```
I think you need whitespace around this

src/main/java/Todo.java
nighoggDatatype: Same as what now? Please clarify in place

src/main/java/Duke.java
```diff
@@ -1,10 +1,124 @@
...
+                duke.add(input);
+                System.out.println("Now you have " + duke.list.size() + " tasks in the list." );
+            } else{
+                throw new DukeWrongCommandException("OOPS!!! I'm sorry, but I don't know what that means :-(");
+            }
+        }
+    } catch (DukeEmptyCommandException e){
+            System.out.println(e.toString());
+        } catch (DukeWrongCommandException w){
+            System.out.println(w.toString());
```
nighoggDatatype:
```suggestion
    } catch (DukeEmptyCommandException e) {
            System.out.println(e.toString());
        } catch (DukeWrongCommandException w) {
            System.out.println(w.toString());
```
I think adding white space is good practice here?

src/main/java/Duke.java
```diff
@@ -1,10 +1,124 @@
...
-        System.out.println("Hello from\n" + logo);
+    /** List of Tasks */
+    private ArrayList<Task> list;
+    private Duke(ArrayList<Task> list){
+
+        this.list = list;
+
+    }
+
+    public void greeting(){
```
nighoggDatatype:
```suggestion
    public void sayGreeting() {
```
I think a verb here is good? Also white space

src/main/java/Duke.java
```diff
@@ -1,10 +1,124 @@
...
+                for(int i = 0; i < len; i++){
+                    int index = i + 1;
+                    Task t = this.list.get(i);
+                    if(t instanceof Todo){
+                        Todo todo = (Todo) t;
+                        System.out.println(index + "." + todo.toString());
+                    } else if ( t instanceof Event){
+                        Event event = (Event) t;
+                        System.out.println(index + "." + event.toString());
+                    } else if ( t instanceof Deadline){
```
nighoggDatatype:
```suggestion
                for (int i = 0; i < len; i++) {
                    int index = i + 1;
                    Task t = this.list.get(i);
                    if (t instanceof Todo) {
                        Todo todo = (Todo) t;
                        System.out.println(index + "." + todo.toString());
                    } else if ( t instanceof Event) {
                        Event event = (Event) t;
                        System.out.println(index + "." + event.toString());
                    } else if ( t instanceof Deadline) {
```
I think you need some white space here

nighoggDatatype (commented):
Need to improve on white spaces, and better function names, but otherwise good, as far as I can see

PR #208, opened 2021-01-27
test

prerthan99:


src/main/java/prerthan/duke/Duke.java
```diff
@@ -0,0 +1,92 @@
...
+				isExit = possibleCommand.get().willTerminate();
+			}
+			catch (DukeInvalidArgumentException | DukeInvalidCommandException | DukeEmptyDetailException
+					| DukeInvalidDateException e)
+			{
+				output.sayError(e);
+			}
+		}
+		return isExit;
+	}
```
dannn-tran: Should you use Egyptian style brackets instead, in line of the module's coding standards?

src/main/java/prerthan/duke/IO/Input.java
```diff
@@ -0,0 +1,167 @@
...
+			case EMPTY:
+				return Optional.empty();
+			case INVALID:
+				throw new DukeInvalidCommandException(commandString,
+						this.getClass().getSimpleName());
+			case BYE:
+				return Optional.of(new ByeCommand(this.tokensWithoutFirst()));
+			default:
+				return Optional.empty();
+		}
```
dannn-tran: Should the `case` clauses have the same indentation as `switch`?

src/main/java/prerthan/duke/Duke.java
```diff
@@ -0,0 +1,92 @@
...
+import sharadhr.duke.io.Output;
+import sharadhr.duke.io.Storage;
+import sharadhr.duke.task.TaskList;
+
+/**
+ *
+ */
+public class Duke
+{
+	public static TaskList tasks;
```
dannn-tran: Shouldn't the indentation be 4 spaces instead of a single tab?

src/test/java/prerthan/duke/io/InputTest.java
```diff
@@ -0,0 +1,57 @@
+package sharadhr.duke.io;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
```
dannn-tran: Shouldn't the imported classes be listed explicitly, instead of using a wildcard?

src/main/java/prerthan/duke/command/Command.java
```diff
@@ -0,0 +1,65 @@
...
+	 * to the command, from the user input.
+	 *
+	 * @param argumentTokens the tokenised input from the user minus the first token
+	 */
+	public Command(String... argumentTokens)
+	{
+		this.argumentTokens = argumentTokens;
+	}
+
+	public static CommandName whichCommand(String token)
```
dannn-tran: Should the method's name be something clearer? For example `getCommandNameFromToken`

dannn-tran (commented):
Looks really good!
Just need to tidy up the code to follow the module's coding standards more closely.

PR #211, opened 2021-01-27
[Kenny Chew] iP

KnitidCeladon23:


src/main/java/Duke.java
```diff
@@ -1,10 +1,147 @@
...
+        System.out.println(farewell);
+        System.out.println(dash);
+        sc.close();
     }
 }
+
+class CheckList {
+    String type;
+    String content;
+    boolean checked;
```
nickyfoo: Perhaps a less ambiguous variable name? If checked means done, then perhaps naming it isDone would be clearer

src/main/java/Duke.java
```diff
@@ -1,10 +1,147 @@
...
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        Scanner sc = new Scanner(System.in);
+        String dash = "____________________________________________________________";
+        String greeting1 = "Pai Kia Bot: Eh harlo! Call me Pai Kia Bot.";
+        String greeting2 = "Pai Kia Bot: What you want?";
+        String greeting3 = "Pai Kia Bot: I will keep a list of your responses ah";
+        String farewell = "Pai Kia Bot: Leave so soon ah? Limpeh sleep first, if got no issue don't disturb me.";
+
+        ArrayList<CheckList> inputList = new ArrayList<>();
```
nickyfoo: Maybe a plural form to represent the collection of Tasks?
Also perhaps a more intuitive name for a Task, instead of a CheckList, which might give the impression that it is a collection of objects?

src/main/java/Duke.java
```diff
@@ -1,10 +1,147 @@
...
+
+        System.out.println(dash);
+        System.out.println("You: " + input);
+        System.out.println(farewell);
+        System.out.println(dash);
+        sc.close();
     }
 }
+
+class CheckList {
```
nickyfoo: See above for name suggestion

nickyfoo (commented):
Overall I like the way you customized the bot, just some little nitpicks about class/variable names

src/main/java/Duke.java
```diff
@@ -1,10 +1,147 @@
...
+
+        System.out.println(dash);
+        System.out.println(greeting1);
+        System.out.println(greeting2);
+        System.out.println(dash);
+        System.out.print("Your response: ");
+        String input = sc.nextLine();
+        System.out.println("");
+
+        while (!input.equals("bye")) {
```
zaiahkoh: I think that using `switch` and `case` would be a lot more readable than `if else` statements here

zaiahkoh: May I also suggest switching to a more OOP-oriented code. Perhaps you could create a Task object and extend it from there for each of the tasks and allow each task to handle its own inputList?

src/main/java/Duke.java
```diff
@@ -1,10 +1,147 @@
...
+                System.out.println(dash);
+            } else if (input.length() > 4 && input.substring(0, 5).equals("done ")) {
+                int index = -1;
+                if (!input.equals("done ")) {
+                    index = Integer.parseInt(input.substring(5));
+                }
+                if (index >= 0 && index <= inputList.size()) {
+                    inputList.set(index - 1, inputList.get(index - 1).makeDone());
+                    System.out.println(dash);
+                    System.out.println("Pai Kia Bot: ok i give this task a done chop: " + inputList.get(index - 1).toString().substring(1));
```
zaiahkoh: Maybe you could break this line into separate lines for readability? I noticed lines longer than 100 chars in several other areas too...

src/main/java/Duke.java
```diff
@@ -1,10 +1,147 @@
...
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        Scanner sc = new Scanner(System.in);
+        String dash = "____________________________________________________________";
+        String greeting1 = "Pai Kia Bot: Eh harlo! Call me Pai Kia Bot.";
+        String greeting2 = "Pai Kia Bot: What you want?";
+        String greeting3 = "Pai Kia Bot: I will keep a list of your responses ah";
+        String farewell = "Pai Kia Bot: Leave so soon ah? Limpeh sleep first, if got no issue don't disturb me.";
```
zaiahkoh: Perhaps these could be formatted as constants instead and be renamed using capital spelling?

src/main/java/Duke.java
```diff
@@ -1,10 +1,147 @@
...
+                if (!input.equals("deadline ")) {
+                    String[] s = input.substring(9).split("/by ");
+                    CheckList c = new CheckList("D", s[0] + "(by: " + s[1] + ")", false);
+                    inputList.add(c);
+                    System.out.println(dash);
+                    System.out.println("Pai Kia Bot: ok i added this task chop chop: " + c.toString().substring(1));
+                    System.out.println(dash);
+                } else {
+                    System.out.println(dash);
+                    System.out.println("Pai Kia Bot: no description leh, try again");
```
zaiahkoh: I think this line `"Pai Kia Bot: no description leh, try again"` could be formatted as a String constant at the start of your code so that it can be reused throughout your program

zaiahkoh (commented):
Overall I think there's quite a bit of nested code and repeated code. I liked the flavour and customization of the bot's personality, and I think the code will improve dramatically with some more OOP abstraction like a separate `Task.java` class.

PR #212, opened 2021-01-27
[Pan Jing Bin] iP

IceBear789:


src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,56 @@
...
+        this.date = deadline;
+    }
+
+    /**
+     * Overloaded constructor to create a deadline task object. It accepts one extra argument
+     * to determine if the task is already done.
+     *
+     * @param taskName The name of the deadline
+     * @param isDone Whether the deadline is already done
+     * @param deadline The date in which the deadline is due
```
NightRaven49: Should these have punctuations?
Possible to review the other Javadocs in the rest of your code too.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,139 @@
...
+        textFile.delete();
+        textFile.createNewFile();
+    }
+
+    /**
+     * Updates the text file whenever the task list changes.
+     *
+     * @param taskList the corresponding task list in which the text file is based on
+     */
+    public static void update(ArrayList<Task> taskList) {
```
NightRaven49: Should taskList be tasks instead?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * Method to read an existing text file and convert it into a corresponding task list within
+     * the Dukebot.
+     *
+     * @param file the text file in which the task list is created from
+     * @param taskList the corresponding task list to copy the text file over
+     * @throws IOException whenever an issue with reading the file arises
+     */
+    public static void convert(File file, ArrayList<Task> taskList) throws IOException {
+        Scanner scan = new Scanner(file);
```
NightRaven49: Perhaps you can consider naming the Scanner object as a noun rather than a verb.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,163 @@
...
+import java.io.IOException;
+import java.time.LocalDate;
+import java.util.ArrayList;
+
+/**
+ * The class that contains the task list. It has operations to add/delete
+ * tasks in the list.
+ */
+class TaskList {
+    private ArrayList<Task> taskList;
```
NightRaven49: Do consider using plural forms for collections.

NightRaven49 (commented):
Overall, code is generally free of errors, though there is some inconsistency in naming and Javadocs. Was easy to read and understand what was the flow in logic, and was particularly impressed with the level of abstraction achieved with the parser. Good job!

src/main/java/duke/Deadline.java
xiinweii98: Strictly speaking, I remember each sentence should end properly with a full stop.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,71 @@
...
+        Scanner scan = new Scanner(System.in);
+        Duke iceBear = new Duke();
+        while (true) {
+            String nextCommand = scan.nextLine();
+            if (nextCommand.equals("bye")) {
+                System.out.println("Bye. Hope to see you again soon!");
+                break;
+            } else {
+                try {
+                    String[] processedText = iceBear.parser.processCommand(nextCommand);
```
xiinweii98: I am pretty sure you can access your private fields instead of going through `iceBear.parser`. However, I am unsure which one you should go with.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,71 @@
...
+            }
+        }
+    }
+    public String[] processCommand(String input) throws DukeException {
+        return this.parser.processCommand(input);
+    }
+
+    public String process(String[] processedInput) {
+        return this.ui.processCommand(processedInput, this.taskList);
+    }
```
xiinweii98: What is the difference between this two methods? I think it would be good to include some java doc comments or make the method names distinct.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,147 @@
...
+            return this.processToDo(input);
+        } else if (command.equals("deadline")) {
+            return this.processDeadLine(input);
+        } else if (command.equals("event")) {
+            return this.processEvent(input);
+        } else if (command.equals("find")) {
+            return this.processFind(input);
+        } else {
+            throw new DukeException("Invalid command");
+        }
```
xiinweii98: A switch case should be apt here.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * The class that deals with loading tasks from the file and saving tasks in the file.
+ */
+class Storage {
+
+    /**
+     * Creates a new text file so that the list of tasks can be saved on the hard disk.
+     */
+    public static void createFile() throws java.io.IOException {
```
xiinweii98: I am not sure why you are using `java.io.IOException` instead of just `IOException`.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,139 @@
...
+                counter++;
+                result[counter] = "";
+            } else if (input.charAt(i) != ' ' || counter >= 2) {
+                result[counter] += input.charAt(i);
+            }
+        }
+        return result;
+    }
+
+    private static LocalDate dateConverter(String date) {
```
xiinweii98: Is this necessary? What if users choose to input dates in another format? So, I would think it is better to make the user input a certain format like yyyy-mm-dd instead of catering to the different formats that a user can input.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,48 @@
...
+        return this.taskName;
+    }
+
+    /**
+     * Takes in a LocalDate object and returns the date of the object in a special format.
+     *
+     * @param date the LocalDate object to print the date of.
+     * @return the string representation of the date of the object in the special format.
+     */
+    public static String printDate(LocalDate date) {
```
xiinweii98, line 69: I think you can use the inbuilt methods to achieve your desired results. [https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/LocalDate.html#format(java.time.format.DateTimeFormatter)](url)

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,163 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.time.LocalDate;
+import java.util.ArrayList;
+
+/**
+ * The class that contains the task list. It has operations to add/delete
+ * tasks in the list.
+ */
+class TaskList {
```
xiinweii98, line 11: It seems like your TaskList is doing the job of Ui returning String objects to be printed/shown to user.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,37 @@
...
+                return taskList.delete(Integer.valueOf(input[1]));
+            } else if (input[0].equals("todo")) {
+                return taskList.addToDo(input[1]);
+            } else if (input[0].equals("deadline")) {
+                return taskList.addDeadline(input[1], LocalDate.parse(input[2]));
+            } else if (input[0].equals("event")) {
+                return taskList.addEvent(input[1], LocalDate.parse(input[2]));
+            } else if (input[0].equals("find")) {
+                return taskList.find(input[1]);
+            }
```
xiinweii98: A switch case would be apt here.

xiinweii98 (commented):
Overall nicely written code. Keep up the good work!

PR #221, opened 2021-01-27
[Goh Jun Wei] iP

JonahhGohh:


src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,92 @@
+package duke.parser;
+
+import duke.command.*;
```
ong6: Imported classes should always be listed explicitly. I noticed the same issue in serval other places too.

src/main/java/duke/storage/DukeFileWriter.java
```diff
@@ -0,0 +1,65 @@
...
+     * @param taskList the current instance of task list used by Duke
+     * @param filePath the file path of the internal storage to save at
+     */
+    public static void saveTaskListInInternalStorage(List<Task> taskList, String filePath) {
+        try {
+            String toBeSavedTaskListString = "";
+            for (int i = 0; i < taskList.size(); i++) {
+                String currentLine = "";
+                Task task = taskList.get(i);
+                if (task instanceof Event) {
```
ong6: Is there a better way to do this instead of using instanceof? Maybe implement a checking system.

src/test/java/duke/parser/ParserTest.java
```diff
@@ -0,0 +1,55 @@
+package duke.parser;
+
+import duke.command.*;
```
ong6: Imported classes should always be listed explicitly.

ong6 (commented):
LGTM. I like your implementations overall except where the nesting was too deep. I noticed a few minor coding standard violations too. Some of the classes are quite redundant. Consider merging them into one class instead.

src/main/java/duke/parser/Parser.java
JonahhGohh: Noted, thanks for the catch!

src/main/java/duke/storage/DukeFileWriter.java
JonahhGohh: Would there be a drawback in using `instanceof`? I implemented it this way because it was convenient.

JonahhGohh:
> Some of the classes are quite redundant. Consider merging them into one class instead.

Hi, thanks for the feedback, do you mind elaborating on which classes are redundant?

src/main/java/duke/storage/DukeFileWriter.java
ong6: To be honest im not too sure, but I think we were taught in 2030 that the use of instanceof breaks the open/close principle in some way you may refer to this StackOverflow page: https://stackoverflow.com/questions/20589590/why-not-use-instanceof-operator-in-oop-design

ong6:
Feel like the extended command classes could have just be done all in one main command class. (Just nitpicking)
Also just spotted that your Task child classes have repeated code in them, you might want to put those in the main Task class instead.

PR #223, opened 2021-01-27
[Zhang Peng] iP

icytornado:


src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
+import java.io.FileNotFoundException;
```
rachelljt: Shouldn't the import statements be arranged in lexicographic order?

src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
...
+        } catch (IOException e) {
+            System.out.println("An error occurred.");
+            e.printStackTrace();
+        }
+
+        //read from the file content into the arrayList at the start.
+        Scanner s = new Scanner(dukeFile);
+        new Storage().loadingFile(arrayList, s);
+
+        System.out.println("--------------------------");
```
rachelljt: I think it is better to store the horizontal lines as a constant as a final variable.

src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
...
+        System.out.println("Hello! I'm Duke");
+        System.out.println("What can I do for you?");
+        System.out.println("--------------------------");
+
+        new Ui().interactWithUser(arrayList, path);
     }
 }
+
```
rachelljt, line 45: I think the empty lines should be deleted to make the code neater.

src/main/java/Parser.java
```diff
@@ -0,0 +1,181 @@
+import java.time.LocalDate;
```
rachelljt, line 1: Should be in lexicographic order.

src/main/java/Parser.java
```diff
@@ -0,0 +1,181 @@
...
+                    throw new DukeException("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+                } else if (input.equals("todo") || input.equals("event") || input.equals("deadline")
+                        || input.equals("done") || input.equals("delete") || input.equals("find")) {
+                    throw new DukeException("☹ OOPS!!! The description of a todo cannot be empty.");
+                }                                // can add in more exceptions
+            } catch (DukeException e) {
+                System.out.println("--------------------------");
+                System.out.println(e.getMessage());
+                System.out.println("--------------------------");
+            }
```
rachelljt: i think its better to have a line separation here

src/main/java/Parser.java
```diff
@@ -0,0 +1,181 @@
...
+
+            } else if (input.contains("find") && !(input.equals("find"))) {
+                System.out.println("--------------------------");
+                System.out.println("Here are the matching tasks in your list:");
+
+                ArrayList<Task> tt = new ArrayList<>();
+
+                int counterOne = 1;
+
+                for (Task ttt :arrayList) {
```
rachelljt: Naming of variable should be meaningful. Perhaps you could change to it to myTask instead, similar to the one above(ArrayList<Task> tt)?

src/main/java/Storage.java
```diff
@@ -0,0 +1,107 @@
+import java.io.FileWriter;
```
rachelljt, line 1: Similar to the one above, it should be arranged in lexicographical order

src/main/java/Storage.java
```diff
@@ -0,0 +1,107 @@
...
+                writeToFile(path, arrayList.get(i) + System.lineSeparator());
+            } else {
+                writeToFile(path, arrayList.get(i) + System.lineSeparator());
+                i++;
+                while (i <= arrayList.size() - 1) {
+                    appendToFile(path, arrayList.get(i) + System.lineSeparator());
+                    i++;  // append to file with in the writetofile.
+                }
+            }
+
```
rachelljt: i think the empty line here should be removed

src/main/java/Storage.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    /**
+     * This is the the method for writing to file
+     * @param path specifies the path to the file
+     * @param s specifies text to add
+     * @return Nothing.
+     */
+
```
rachelljt, line 109: I think there shouldn't be an empty line here

src/main/java/TaskTest.java
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * This is the the test method for markAsDone method.
+     * @return Nothing.
+     */
+    @Test
+    void markAsDone() {
+
+        task.markAsDone();
+        if (task.isDone = true) {
+            checker =1;
```
rachelljt: Should this be written as checker = 1 instead?

rachelljt (commented):
I think its pretty well done, just a few formatting errors to be fixed!

src/main/java/Duke.java
icytornado: Thank you.

src/main/java/Duke.java
icytornado: got it. thank you.

src/main/java/Parser.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /**
+     * This is the the method for making sense of the user commands
+     * @param arrayList takes in the arrayList
+     * @param path specifies the path to the file
+     * @param input takes in the a scanner to scan
+     * @return Nothing.
+     */
+    public String makingSenseOfUserCommand(ArrayList<Task> arrayList, String path, String input) {
```
edelinetenges, line 21: Would it be better to name this method "makeSenseOfUserCommand" instead? Other than that, I like how the method name clearly explains what the method does!

src/main/java/Parser.java
```diff
@@ -0,0 +1,170 @@
...
+           String holder;
```
edelinetenges: Perhaps naming this String "output" could be more meaningful?

src/main/java/Parser.java
```diff
@@ -0,0 +1,170 @@
...
+            } else if (input.equals("list")) {
+                holder = "Here are the tasks in your list:" + "\n";
+                for (Task m : arrayList) {
+                    holder = holder.concat(m.index + ". " + m);
+                    holder = holder.concat("\n");
+                }
+
+
+            } else if (input.contains("done ")) {
+                int x = Integer.parseInt(input.substring(5));
```
edelinetenges: I think naming this variable "taskIndex" or something like that could help improve readability? Especially since you will keep using it repeatedly in the else block below.

src/main/java/Parser.java
```diff
@@ -0,0 +1,170 @@
...
+                holder = holder.concat("Now you have " + arrayList.size() + " task(s) in the list");
+
+
+
+            } else if (input.contains("deadline")) {
+
+                holder = "Got it. I've added this task: "+ "\n";
+                String[] parts = input.split("/",2);
+                String part1 = parts[0];
+                String part2 = parts[1];
```
edelinetenges: Maybe it would be clearer to mention in the variable name what parts these are?

For example:
"inputParts" instead of just "parts", 
"inputFirstPart" instead of "part1", 
"inputSecondPart" instead of "part2"

src/main/java/Parser.java
```diff
@@ -0,0 +1,170 @@
...
+
+                //e.g. deadline return book /by 02/12/2019
+                if (part2.contains("/")) {
+                    String dateString = part2.substring(3);
+                    String temp = dateString.substring(6)+"-"+dateString.substring(3,5) + "-" +dateString.substring(0,2);
+                    LocalDate xx = LocalDate.parse(temp);
+                    String f = xx.format(DateTimeFormatter.ofPattern("MMM d yyyy"));
```
edelinetenges: Perhaps replacing "xx" with "localDate" and then replacing "f" with "formattedDateString" could improve readability?

src/main/java/Parser.java
```diff
@@ -0,0 +1,170 @@
...
+
+                    }
+                } catch (DukeException e) {
+                    return e.getMessage();
+                }
+
+            } else if (input.contains("find ")) {
+                holder = "Here are the matching tasks in your list: \n";
+                ArrayList<Task> tt = new ArrayList<>();
+                int counterOne = 1;
```
edelinetenges: I noticed that you are using 2 counters so you named them counterOne and counterTwo. I like the fact that you try to differentiate them, but maybe it would be more intuitive to name them something else?

src/main/java/Storage.java
```diff
@@ -0,0 +1,107 @@
...
+public class Storage {
+    public Storage() {
+    }
+    /**
+     * This is the method to load file from disk
+     * @param arrayList takes in the arrayList
+     * @param s takes in the a scanner to scan
+     * @return Nothing.
+     */
+    public void loadingFile(ArrayList<Task> arrayList, Scanner s) {
```
edelinetenges, line 19: Would it be better to name this "loadFile" instead?

src/main/java/Storage.java
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+    }
+    /**
+     * This is the the method for saving changes to the disk
+     * @param arrayList takes in the arrayList
+     * @param path specifies to file path to write to
+     * @return Nothing.
+     */
+    public void savingFile(ArrayList<Task> arrayList, String path) {
```
edelinetenges, line 73: Similarly, would it be better to name this "saveFile" instead?

edelinetenges (commented):
Overall good coding quality, just a few adjustments to naming that could help improve readability!

PR #226, opened 2021-01-27
[Heng Yi Qun] iP

hengyiqun:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Deadline is a Task object that is define by a due date
+ */
+public class Deadline extends Task implements DueDate {
+    protected LocalDate time;
+
+    /**
+     * Constructs a deadline that has yet to be completed
+     * @param description description of the deadline
+     * @param time due date in yyyy-MM-dd format (e.g. 2021-03-12)
```
wangtao0717: According to code standard, I think leaving a line break between the description and parameter section will be better.

src/main/java/Duke.java
```diff
@@ -1,10 +1,291 @@
...
+        }
+    }
+
+    public void run() {
+        while (!this.ui.shouldExit() && this.ui.hasNext()) {
+            this.tasks = ui.process(this.tasks);
+            try {
+                FileWriter fw = new FileWriter(this.filePath);
+                for (Task t : this.tasks.getList()) {
+                    String rmb = "";
```
wangtao0717: Perhaps a more intuitive variable name here?

src/main/java/Duke.java
```diff
@@ -1,10 +1,291 @@
...
+                        } else {
+                            taskDescription += parts[i];
+                        }
+                        first = false;
+                    }
+                    Todo todo = new Todo(taskDescription);
+                    store.add(todo);
+                    System.out.println("     " + todo);
+                    System.out.println("     Now you have " + store.size() + " tasks in the list.");
+                } else if (parts[0].equals("deadline")) {
```
wangtao0717: Perhaps it may be more concise to replace if-else if method with switch method here.

src/main/java/Duke.java
```diff
@@ -1,10 +1,291 @@
...
+            } catch(WrongArgumentException e) {
+                System.out.println(e.getMessage());
+                System.out.println("    ____________________________________________________________");
+            } catch (InsufficientArgumentsException e) {
+                System.out.println(e.getMessage());
+                System.out.println("    ____________________________________________________________");
+            }
+        }
+    }*/
+}
```
wangtao0717: The Duke.java seems too long, will it be better to extracte some code out as some classes or functions?

src/main/java/Parser.java
```diff
@@ -0,0 +1,190 @@
...
+    public Parser parse(String input) {
+        try {
+            String[] parts = input.split(" ");
+            if (parts[0].equals("bye")) {
+                return new Parser("bye", null, null, -1);
+            } else if (parts[0].equals("list")) {
+                return new Parser("list", null, null, -1);
+            } else if (parts[0].equals("done")) {
+                if (parts.length == 1) {
+                    throw new InsufficientArgumentsException("     ☹ OOPS!!! The " +
```
wangtao0717: Perhaps it is better to break before an operator "+"

src/main/java/Ui.java
```diff
@@ -0,0 +1,147 @@
...
+            System.out.println(msg);
+            this.exit = true;
+            return tasks;
+        } else if (parser.getTaskType().equals("list")) {
+            String msg = "     Here are the tasks in your list:\n";
+            ArrayList<Task> taskList = tasks.getList();
+            int count = 1;
+            for (Task t : taskList) {
+                if (count == taskList.size()) {
+                    msg += "     " + count + "." + t.toString();
```
wangtao0717: There is a lot of the same indentation, maybe it can be extracted into a variable, which can be changed easily later？

wangtao0717 (commented):
In general, there is no major problem in code quality. The code is clear and easy to understand. If it can replace some long if-elseif methods with the switch method, the code may be more concise. And I think it would be better if some parts are not repeated as often, which can be extracted out to some classes or methods.

src/main/java/Duke.java
jasaaanlim: Even though 'rmb' is just a temporary short-lived variable, would a more readable variable name make things easier for other collaborators and when looking back in future?

src/main/java/Duke.java
```diff
@@ -1,10 +1,291 @@
...
+                        rmb += "D";
+                    } else if (t instanceof Todo) {
+                        rmb += "T";
+                    }
+                    rmb += " | ";
+                    if (t.isCompleted()) {
+                        rmb += "1";
+                    } else {
+                        rmb += "0";
+                    }
```
jasaaanlim: Could a switch case be a neater alternative as compared to a long list of if-else?

src/main/java/Duke.java
```diff
@@ -1,10 +1,291 @@
...
+                FileWriter fw = new FileWriter(this.filePath);
+                for (Task t : this.tasks.getList()) {
+                    String rmb = "";
+                    if (t instanceof Event) {
+                        rmb += "E";
+                    } else if (t instanceof Deadline) {
+                        rmb += "D";
+                    } else if (t instanceof Todo) {
+                        rmb += "T";
+                    }
```
jasaaanlim: I noticed a few repetitions throughout the various classes where you had to use multiple if else statements to display the event type. Have you thought about handling the display of an event type in the respective toString() methods of the Task? Would that be a more efficient way?

src/main/java/Duke.java
```diff
@@ -1,10 +1,291 @@
...
+        Scanner sc = new Scanner(System.in);
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+        String filePath = "data/duke.txt";
+        try {
+            File file = new File(filePath);
+            Scanner fileScanner = new Scanner(file);
+            while (fileScanner.hasNext()) {
+                String curRecTask = fileScanner.nextLine();
+                String[] items = curRecTask.split(" \\| ");
+                boolean done = (items[1].equals("1") ? true : false);
```
jasaaanlim: Noticed that a few of your boolean variables were not named to sound like boolean variables. E.g. the 'first' and 'done' booleans. In this case, would it be more intuitive to name this boolean variable as isDone instead?

src/main/java/Parser.java
```diff
@@ -0,0 +1,190 @@
...
+                    if (i != parts.length - 1) {
+                        toFind += parts[i] + " ";
+                    } else {
+                        toFind += parts[i];
+                    }
+                }
+                return new Parser("find", toFind, null, -1);
+            } else if (parts[0].equals("todo")) {
+                if (parts.length == 1) {
+                    throw new InsufficientArgumentsException("     ☹ OOPS!!! The " +
```
jasaaanlim: Is there a reason why you are hardcoding the tabs? Would '\t' may be a neater way of displaying your tabs?

src/main/java/Duke.java
jasaaanlim: I agree with @wangtao0717 on the Duke.java being a little too long, maybe you could look into that.

src/main/java/Ui.java
```diff
@@ -0,0 +1,147 @@
...
+            Todo todo = new Todo(taskDescription);
+            tasks = tasks.add(todo);
+            msg += "     " + todo + "\n";
+            msg += "     Now you have " + tasks.size() + " tasks in the list.";
+            System.out.println(this.format(msg));
+        } else if (parser.getTaskType().equals("deadline")) {
+            String msg = "     Got it. I've added this task:\n";
+            LocalDate dueDate = parser.getDueDate();
+            String taskDescription = parser.getDescription();
+            Deadline deadline = new Deadline(taskDescription, dueDate);
```
jasaaanlim: Since you do not plan to use dueDate or taskDescription, could you directly place parser.getDueDate() and parser.getDescription() in the constructor instead?

jasaaanlim (commented):
The program is generally well written with some minor issues with the naming conventions and there can be some improvements on reusing to code and variable reuse.

src/main/java/DueDate.java
```diff
@@ -0,0 +1,6 @@
+/**
+ * DueDate is an interface for inheritance by tasks with a due date
+ */
+public interface DueDate {
```
tlylt: Good job on naming the interface correctly in PascalCase. I like your use of an interface here.

src/main/java/Duke.java
```diff
@@ -1,10 +1,291 @@
+import java.io.IOException;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+
+/**
+ * Duke allows the user to maintain a list of tasks
+ */
 public class Duke {
+
```
tlylt: Hi, just a small inconsistency that I spotted. For some of your classes, there is an empty line between the class declaration and the class properties. However, some of the classes don't. 

Perhaps it will be more consistent if all classes have an empty line in between or none of them have an empty line in between?

src/main/java/Duke.java
```diff
@@ -1,10 +1,291 @@
...
+                    store.add(deadline);
+                    System.out.println("     " + deadline);
+                    System.out.println("     Now you have " + store.size() + " tasks in the list.");
+                } else if (parts[0].equals("event")) {
+                    if (parts.length == 1) {
+                        throw new InsufficientArgumentsException("     ☹ OOPS!!! The " +
+                                "description of an event cannot be empty.");
+                    }
+                    System.out.println("     Got it. I've added this task:");
+                    boolean flag = false, first = true;
```
tlylt: Perhaps a more descriptive name for the boolean variable (instead of "flag") will be better?

tlylt (commented):
Overall good code! Just a few nits to fix.

PR #229, opened 2021-01-27
[Li Gang] iP

whatthelump:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,16 @@
+package main.java;
+
+public class Deadline extends Task {
```
dvdweien: perhaps the deadline class does not need to be public? As deadline class is only accessed within the package. I think some of the other classes do not have to be public as well.

src/main/java/DukeException.java
```diff
@@ -0,0 +1,14 @@
+package main.java;
+
+public class DukeException extends Exception {
```
dvdweien: Perhaps DukeException does not need to be public? As DukeException would not be used outside of the package, I think the other exceptions do not need to be public as well

src/main/java/Duke.java
```diff
@@ -1,10 +1,110 @@
+package main.java;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+
```
dvdweien: perhaps a javadoc might be needed as it is a public class

src/main/java/Task.java
```diff
@@ -0,0 +1,29 @@
...
+
+public class Task {
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
```
dvdweien: Perhaps there should be javadoc since it is a public method?

dvdweien (commented):
Overall I found your code to have good names and could not find many coding style violations.

src/main/java/todo.java
```diff
@@ -0,0 +1,16 @@
+package main.java;
+
+public class todo extends Task {
+
+    protected String by;
```
lrj689: Consider using a clearer name, ie deadline, date

src/main/java/Duke.java
```diff
@@ -1,10 +1,110 @@
...
+        System.out.println("Greetings and salutations! I am Duke, a helpful chatbot.\nPlease enter your command(s).");
+        System.out.println("Available commands: todo (task), deadline (task) /by (time), event (task) /at (time)");
+        System.out.println("To view the list of tasks, enter the command 'list'.");
+        System.out.println("To exit, enter the command 'bye'.");
+        Scanner sc = new Scanner(System.in);
+        List<Task> list = new ArrayList<>(100);
+        while (true) {
+            String input = sc.nextLine();
+            try {
+                if (input.equals("bye")) {
```
lrj689: Consider trying out switch case statements

src/main/java/Event.java
```diff
@@ -0,0 +1,17 @@
+package main.java;
+
+public class Event extends Task {
+
+    protected String at;
```
lrj689: Consider using a clearer name

lrj689 (commented):
Looks good! Keep it up

PR #230, opened 2021-01-27
[Wang Yuchen] iP

w-yuchen:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,16 @@
...
+    private LocalDate deadline;
+    private DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-M-dd");
+    Deadline(String name, String deadLine) {
+        super(name);
+        this.deadline = LocalDate.parse(deadLine, formatter);
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + this.deadline.format(DateTimeFormatter.ofPattern("MMM D YYYY")) + ")";
```
DrWala: Perhaps the deadline formatting can be abstracted out to a getter, which would make this bit look cleaner. This issue exists in some other parts of the code base too.

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+        while (true) {
+            raw_in = sc.nextLine().trim();
+            input = raw_in.split(" ");
+            System.out.println(h_rule);
+            String joined = "";
+            String timing = "";
+            if (raw_in.equals("bye")) {
+                System.out.println("Goodbye! Have a nice day!\n");
+                break;
+            } else {
```
DrWala, line 122: It might be helpful to remove the else clause so that the core logic is not so deeply nested, and the happy path is more prominent

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+                switch (input[0]) {
```
DrWala: I believe based on the style guides, the case statements should not be indented. Perhaps double-check this?

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+                    case "list":
+                        int index = 1;
+                        for (Task s : itemList) {
```
DrWala: Is there a particular reason you are using the enhanced for loop here? I see that you have an index variable, maybe a normal for loop would then suffice, as the iterator is contained within the loop's scope

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+                        for (int i = 1; i < input.length; i++) {
+                            joined = joined + " " + input[i];
+                        }
+                        itemList.add(new Todo(joined));
+                        fileContents.add(new Todo(joined).toString());
+                        count++;
+                        System.out.println("Added " + raw_in + "\nYou now have " + count + " items in your list.");
+                        break;
+                    case "deadline":
+                        int seq = 0;
```
DrWala: Perhaps `seq` can be a more intuitive name, or a comment explaining what it does would be helpful

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+            input = raw_in.split(" ");
+            System.out.println(h_rule);
+            String joined = "";
+            String timing = "";
+            if (raw_in.equals("bye")) {
+                System.out.println("Goodbye! Have a nice day!\n");
+                break;
+            } else {
+                switch (input[0]) {
+                    case "list":
```
DrWala: It might be easier to maintain if the case constants are extracted out somewhere else, either as an enum or some constant variables

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+                        break;
+                    case "done":
+                        int itemToBeUpdatedIndex = Integer.parseInt(input[1]) - 1;
+                        Task task = itemList.get(itemToBeUpdatedIndex);
+                        task.markDone();
+                        fileContents.set(itemToBeUpdatedIndex, task.toString());
+                        System.out.println("Alright, I will mark this as done.\n" + input[1] + ". " + task.toString());
+                        break;
+                    case "todo":
+                        for (int i = 1; i < input.length; i++) {
```
DrWala: This might be able to be abstracted out using the `String.join` function.

This kind of abstraction may be applicable in a few other parts of the code base too.

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+                            joined = joined + " " + input[i];
+                        }
+                        itemList.add(new Todo(joined));
```
DrWala: I see that itemList and fileContents are frequently modified in the same way (addition/update/removal) together, maybe abstracting out to a function would be cleaner and easier to maintain

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
 public class Duke {
+    public static String h_rule = "────────────────────────────────────────────────────────────────────";
+
+    public static int dukeRunner(String log) throws DukeNotFoundException, DukeDescriptionException, DukeTimingException, FileNotFoundException {
+        Scanner sc = new Scanner(System.in);
+
+        String raw_in;
+        String[] input;
+        List<Task> itemList = new ArrayList<>();
+        int count = 0;
```
DrWala: Perhaps it might be more future-proof to use `itemList.count()` instead of manually maintaining a count value. Is there a reason you did it this way?

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+        List<String> fileContents = null;
+
+        try {
```
DrWala: Perhaps all the file-based IO can be abstracted away to another class for easier setup/teardown and CRUD operations. Is there a reason to keep it purely within the main class?

DrWala (commented):
In general, I think a number of functions can be abstracted out to their own units instead of being in the main class. It would probably help the maintainability and testability of the code too.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,16 @@
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task {
+    private LocalDate deadline;
+    private DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-M-dd");
+    Deadline(String name, String deadLine) {
```
vevek: Perhaps deadline in the constructor parameter be written as "deadline" instead of deadLine.

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
+import java.io.*;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+
 public class Duke {
+    public static String h_rule = "────────────────────────────────────────────────────────────────────";
```
vevek: Perhaps h_rule can be renamed as HORIZONTAL_LINE_RULE. Perhaps it can be made a constant and hence named in all caps.

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+
+    public static int dukeRunner(String log) throws DukeNotFoundException, DukeDescriptionException, DukeTimingException, FileNotFoundException {
+        Scanner sc = new Scanner(System.in);
+
+        String raw_in;
```
vevek: Should raw_in be named as rawIn or rawInput instead?

src/main/java/Duke.java
vevek: As mentioned by DrWala, perhaps "seq" can be renamed. What does "seq" stand for over here?

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+                        itemList.add(new Todo(joined));
+                        fileContents.add(new Todo(joined).toString());
+                        count++;
+                        System.out.println("Added " + raw_in + "\nYou now have " + count + " items in your list.");
+                        break;
+                    case "deadline":
+                        int seq = 0;
+                        while (!input[seq].equals("/by")) {
+                            joined = joined + " " + input[seq];
+                            seq++;
```
vevek: Perhaps a gap between the statement seq++ and the if statement would make it easier to read. I noticed this elsewhere too.

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+                            timing = timing + " " + input[i];
+                        }
+
+                        itemList.add(new Deadline(joined, timing.trim()));
+                        fileContents.add(new Deadline(joined, timing.trim()).toString());
+                        count++;
+                        System.out.println("Added " + joined + "\nYou now have " + count + " items in your list.");
+                        break;
+                    case "event":
+                        int seq2 = 0;
```
vevek: Perhaps "seq2" here can be renamed to secondSequence as it would make it easier to distinguish from the above statements.

src/main/java/Duke.java
vevek: As mentioned by DrWala, I think case statements should not be indented, but rather inline with the switch statement. Please check again. There's a setting in the IDE to make it easier for you to auto format in the future too!

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+
+        System.out.println("Greetings! I am Duke! How may I assist you?\n" + h_rule);
+        int res = -1;
```
vevek: Should "res" be named as "result" over here? It might make it easier to read.

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+        File logFile = new File("./logs");
+
+        try {
+            if (logFile.isFile()) {
+                Scanner logs = new Scanner(logFile);
+                while (logs.hasNextLine()) {
```
vevek: Would it be better to have a gap between the scanner statement and the while loop statement? It might make it easier to read. I noticed this in a few other places too.

vevek (commented):
Hi Yuchen! I found the overall code very pleasant to go through. There were just a few minor things that could be improved such as having switch statement's cases inline as well as making some names more easily understandable.

src/main/java/Deadline.java
Cp-John: Put a blank line to separate constructor from the private fields according to Java coding standard. And I think making the constructor public is better?

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
+import java.io.*;
```
Cp-John: Import classes explicitly according to Java coding standard.

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+
 public class Duke {
+    public static String h_rule = "────────────────────────────────────────────────────────────────────";
+
+    public static int dukeRunner(String log) throws DukeNotFoundException, DukeDescriptionException, DukeTimingException, FileNotFoundException {
```
Cp-John: Since there are some many different types of exceptions the method may throw, maybe it is better to create a parent class DukeException of all of them and the method header may be shorter. e.g. `public static int dukeRunner(String log) throws DukeException {`

src/main/java/Duke.java
Cp-John: The body of each case part is so long.  I think you may put them into methods which may increase readability of your code.

Cp-John (commented):
LGTM. But remember to follow the Java coding standard. And if a method is too long, can abstract part of the method into another method with a rational name which largely increases readability of your code.

PR #234, opened 2021-01-28
[Jacob-109] iP

Jacob-109:
The intended commits are those done on 21Jan 2021, i wanted to edit Level 3 but in the end it duplicated the commits above as well. I should have fixed the issue already. Sorry for the inconvenience caused if you saw the extra commits.


src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
+public class Deadline extends Task {
```
litone01: Should the class be added or grouped inside a package? This applies to other classes as well.

src/main/java/Duke.java
```diff
@@ -1,10 +1,75 @@
...
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+        Scanner sc = new Scanner(System.in);
+        ArrayList<Task> list = new ArrayList<>();
+
+        while(sc.hasNext()) {
```
litone01: Will it be better to abstract and break this method into many smaller units? It will help in the future also.

src/main/java/Duke.java
```diff
@@ -1,10 +1,75 @@
+import java.lang.reflect.Array;
+import java.util.Scanner;
+import java.util.ArrayList;
+
 public class Duke {
```
litone01: Will it be better to add javadoc to all public classes and methods? The same applies to all other public classes and methods.

src/main/java/Duke.java
```diff
@@ -1,10 +1,75 @@
+import java.lang.reflect.Array;
```
litone01: Can I check if this import is being used? Because Array should be present by default in Java.

litone01 (commented):
Keep it up Jacob! You can do it:)

src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
...
+
+    protected String by;
```
chewwh09: You could have a better convention such as dateTime. "by" alone doesn't have any meaning.

src/main/java/Duke.java
```diff
@@ -1,10 +1,75 @@
...
-    public static void main(String[] args) {
+    public static void main(String[] args) throws DukeException{
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+        Scanner sc = new Scanner(System.in);
+        ArrayList<Task> list = new ArrayList<>();
```
chewwh09: Should be lists instead of list. Also, the naming convention could have been better, perhaps using tasks is better? After all, your code is being read by many people with different backgrounds. So it would be better if your naming convention is clearer.

src/main/java/Event.java
```diff
@@ -0,0 +1,14 @@
+public class Event extends Task {
+
+    protected String datetime;
```
chewwh09: Your deadline by naming convention should have followed this format as well.

src/main/java/Duke.java
```diff
@@ -1,10 +1,75 @@
...
+
+        while(sc.hasNext()) {
+            String word = sc.nextLine();
+            if (word.equals("bye")) {
```
chewwh09: You can change the if-else to switch case for better readability.

src/main/java/Duke.java
```diff
@@ -1,10 +1,75 @@
...
+                System.out.println("Bye, hope to see you again!");
+                break;
+            } else if (word.equals("list")) {
+                int size = list.size();
+                System.out.println("Here are the tasks in your list:");
+                for (int i = 1; i <= size; i++) {
```
chewwh09: Could have just do for (int i = 1; i <= list.size(); i++)

src/main/java/Duke.java
```diff
@@ -1,10 +1,75 @@
+import java.lang.reflect.Array;
+import java.util.Scanner;
+import java.util.ArrayList;
+
 public class Duke {
-    public static void main(String[] args) {
+    public static void main(String[] args) throws DukeException{
```
chewwh09: You should catch the DukeException, if not your program would stop running due to the exception.

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
```
chewwh09: It seems like getStatusIcon() is used in this task class only, hence you can change public to private.

chewwh09 (commented):
Name convention could have been better overall and some nitpicking.

src/main/java/Deadline.java
litone01: Hi! Thanks for the comment, here I did not implement "by" as a datetime object for deadline:) Thus, it is kept as a string. However, I may change the name of the variable.

src/main/java/Duke.java
litone01: Alright!

src/main/java/Deadline.java
litone01: so sorry I thought this was my iP, please ignore my replies if any:)

PR #235, opened 2021-01-28
Michael Ong iP

maikongeh:
pull request for week 3


src/main/java/Duke.java
```diff
@@ -1,10 +1,261 @@
...
+        }
+        catch (FileNotFoundException e){
+            throw new FileNotFoundException("No File Detected");
+        }
+        System.out.println("____________________________________");
+        System.out.println("Hello! I'm Duke\nWhat can I do for you?");
+        System.out.println("____________________________________");
+
+        Scanner input = new Scanner(System.in);
+        while(input.hasNextLine()) {
```
tashawan23: Maybe it would be better if you avoid deep nesting to improve code quality

src/main/java/Duke.java
```diff
@@ -1,10 +1,261 @@
...
+
+        }
     }
+
+    private static ArrayList<Task> readInput(ArrayList<String> strings) {
+        ArrayList<Task> tasks = new ArrayList<>();
+        for(String str: strings) {
+            char identifier = str.charAt(1);
+            switch(identifier) {
+                case 'D':
```
tashawan23: Maybe it would be better to leave out the indentation for case clauses

src/main/java/Duke.java
```diff
@@ -1,10 +1,261 @@
...
+        ArrayList<String> listOfTasks;
+        ArrayList<Task> myList;
+       try {
+            listOfTasks = database.readFile();
+            myList = readInput(listOfTasks);
+        }
+        catch (FileNotFoundException e){
+            throw new FileNotFoundException("No File Detected");
+        }
+        System.out.println("____________________________________");
```
tashawan23: Would it be better to have this line format as a constant since it is repeated several times in the class?

src/main/java/Database.java
```diff
@@ -0,0 +1,47 @@
...
+
+    String name;
+
+
+    public Database(String name){
+        this.name = name;
+    }
+
+    public ArrayList<String> readFile() throws FileNotFoundException {
+        try {
```
tashawan23: I like your indentations in this class!

tashawan23 (commented):
Overall, I found your code quite easy to understand for the most part except for a few places where nesting was too deep. I have noted some repeated code and coding standard violations. :)

src/main/java/Database.java
```diff
@@ -0,0 +1,47 @@
...
+            while (s.hasNext()) {
+                tasksStringName.add(s.nextLine());
+            }
+            return tasksStringName;
+        } catch (FileNotFoundException e) {
+            throw new FileNotFoundException("File not found");
+        }
+    }
+
+    public  void writeTaskToFile(List<Task> tasks){
```
pyuxiang: Be careful of the spaces 😉

```suggestion
    public void writeTaskToFile(List<Task> tasks){
```

src/main/java/Database.java
```diff
@@ -0,0 +1,47 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+
+public class Database {
+
+    String name;
+
+
```
pyuxiang: You might have inadvertently added an extra blank line here:
https://se-education.org/guides/conventions/java/intermediate.html#layout
```suggestion
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,261 @@
...
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+
+    public static void main(String[] args) throws DukeException, FileNotFoundException {
+        Database database = new Database( "data.txt");
+        ArrayList<String> listOfTasks;
+        ArrayList<Task> myList;
+       try {
```
pyuxiang: The indentation is a little off here: should be in line with the block.
```suggestion
        try {
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,261 @@
...
+            listOfTasks = database.readFile();
+            myList = readInput(listOfTasks);
+        }
+        catch (FileNotFoundException e){
```
pyuxiang: The ``catch`` keyword should be in line with the closing brace.

```suggestion
        } catch (FileNotFoundException e){
```

src/main/java/Duke.java
pyuxiang: Another space missed 😉 

```suggestion
        while (input.hasNextLine()) {
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,261 @@
...
+        System.out.println("____________________________________");
+        System.out.println("Hello! I'm Duke\nWhat can I do for you?");
+        System.out.println("____________________________________");
+
+        Scanner input = new Scanner(System.in);
+        while(input.hasNextLine()) {
+            String s = input.next();
+            switch (s){
+                case "todo":
+                    System.out.println("____________________________________");
```
pyuxiang: The ``case`` statement should be in line with the ``switch``, and the case blocks one indentation lower.
```suggestion
            switch (s){
            case "todo":
                System.out.println("____________________________________");
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,261 @@
...
+                    try {
+                        String s1 = input.nextLine();
+                        if (s1.equals("")) {
+                            throw new DukeException("Enter a valid todo");
+                        } else {
+                            char[] chars = s1.toCharArray();
+                            if(chars[1] == ' '){
```
pyuxiang: Another missing space 😉

src/main/java/Duke.java
```diff
@@ -1,10 +1,261 @@
...
+                                throw new DukeException("Enter valid todo");
+                            } else {
+                                try{
```
pyuxiang: And another 😉

src/main/java/Duke.java
```diff
@@ -1,10 +1,261 @@
...
+        System.out.println("____________________________________");
+
+        Scanner input = new Scanner(System.in);
+        while(input.hasNextLine()) {
+            String s = input.next();
+            switch (s){
+                case "todo":
+                    System.out.println("____________________________________");
+                    try {
+                        String s1 = input.nextLine();
```
pyuxiang: You might have missed this: variable names could do with a more specific name, especially since the scope isn't small. Perhaps `s1` can be renamed as `userInputLine`.

src/main/java/Duke.java
pyuxiang: Applies to a couple other lines below.

src/main/java/Duke.java
```diff
@@ -1,10 +1,261 @@
...
+                                    String deadline = parts[1];
+                                    System.out.println("Got it. I've added this deadline:");
+                                    Deadline newDeadline = new Deadline(deadlineDesc, deadline);
+                                    System.out.println(" " + newDeadline.toString());
+                                    myList.add(newDeadline);
+                                    System.out.println("Now you have " + myList.size() + " tasks in the list.");
+                                    database.writeTaskToFile(myList);
+
+                                }
+                            } catch(DukeException e) {
```
pyuxiang: Not a coding standard, but figured I could chip in. Since the `catch` block is the same as that below in line 86, you can consider removing the `try-except` block here.

src/main/java/Duke.java
```diff
@@ -1,10 +1,261 @@
...
+                    } catch (DukeException e){
+                        System.out.println(e.getMessage());
+                    }
+                    System.out.println("____________________________________");
+                    break;
+
+                case "list":
+                    System.out.println("____________________________________");
+                    System.out.println("Here are the tasks in your list:");
+                    for(int i=1; i< myList.size()+1; i++){
```
pyuxiang: Note the spaces here as well:
```suggestion
                    for (int i = 1; i < myList.size() + 1; i++) {
```

src/main/java/Database.java
```diff
@@ -0,0 +1,47 @@
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+
+public class Database {
```
pyuxiang: Consider adding a couple of class-level comments describing the class.
Might not be immediately clear otherwise what this `Database` class is intended for.

src/main/java/Database.java
```diff
@@ -0,0 +1,47 @@
+import java.io.File;
```
pyuxiang: Your `import` statements are very nicely formatted 🎉

pyuxiang (commented):
A couple of code standard comments, but otherwise adheres to standards pretty nicely!

PR #238, opened 2021-01-28
[Galen Cheung] iP

cheunggalen:


src/main/java/Event.java
```diff
@@ -0,0 +1,93 @@
...
+                }
+            }
+            end += combinedHour.toString() + ":";
+        } else {
+            if (onesHour.equals(0)) {
+                end += "12:";
+            } else {
+                end += onesHour.toString() + ":";
+            }
+        }
```
noelmathewisaac: Perhaps you can consider using else if and else statements here instead of a nested if-else to improve code readability?  I noticed the same in other places too.

src/main/java/Duke.java
```diff
@@ -1,10 +1,53 @@
...
+    }
+
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        new Duke("./data/duke.txt").run();
```
noelmathewisaac, line 51: I like how you used a relative path here 👍

noelmathewisaac: I like how you used a relative path here 👍

src/main/java/Ui.java
```diff
@@ -0,0 +1,37 @@
...
+        System.out.println(response);
+    }
+
+    /*
+     * Replace the current response with a new one.
+     * Respond with the new response.
+     *
+     * @param statement The new response that will replace the current one.
+     */
+    public void replace(String statement) {
```
noelmathewisaac: Should this function be renamed to 'replaceResponse' or 'newResponse' to make the intent of the function clearer? I wasn't sure what 'replace; did when I encountered the function in the Parser.java file.

src/main/java/Parser.java
```diff
@@ -0,0 +1,231 @@
...
+                }
+                ui.replace("Got it. I've added this task:");
+                String description = "";
+                for (int i = 1; i < length; i++) {
+                    if (i + 1 == length) {
+                        description += line[i];
+                    } else {
+                        description += line[i] + " ";
+                    }
+                }
```
noelmathewisaac: Should the logic used extract out the description be simplified to make it more understandable? Perhaps instead of splitting on spaces and combing all the words after todo, you can simply remove the word 'todo' from the String? This will make require only a single line of code and is much more compact.

Something like:

```
String description  = line.replace("todo", "").strip()  //.strip() is to remove trailing spaces
```

The same can be applied in other parts of your code for descriptions of deadlines and events.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,63 @@
...
+        return deletedTask;
+    }
+
+    /*
+     * Replace a specific task with another in the list.
+     *
+     * @param index Position of the to-be-replaced task in the list.
+     * @oaram task The new task that will replace the current one.
+     */
+    public void replaceTask(int index, Task task) {
```
noelmathewisaac: I like the intuitive naming of all the functions in this class.

src/main/java/Storage.java
```diff
@@ -0,0 +1,101 @@
...
+            String timeWithBracket = "";
+            int twbLength = 0;
+            int twbLastIndex = 0;
+            String year = "";
+            String yearLine = "";
+            int descLastIndex = 0;
+            Task task = null;
+            String description = "";
+            String date = "";
+            String time = "";
```
noelmathewisaac: Perhaps you can consider abstracting away the formatting/extraction of dates/times by creating a separate class for functions related to these? There are a lot of variables declared here and abstracting these details will make the code easier to understand.

src/main/java/Task.java
```diff
@@ -0,0 +1,141 @@
...
+        Integer onesMin = cArr[3] - '0';
+        String formattedTime = "";
+        boolean isAfternoon = false;
+        if (!tensHour.equals(0)) {
+            Integer combinedHour = tensHour * 10 + onesHour;
+            if (combinedHour >= 12) {
+                isAfternoon = true;
+                if (combinedHour >= 13) {
+                    combinedHour -= 12;
+                }
```
noelmathewisaac: Should you leverage the LocalDate or some other library to convert the date and time format? I feel that doing so will make the code less cluttered and save you the trouble of writing everything from scratch. This can also be applied to all the classes which inherit from Task.  

Maybe consider something along the lines of this [tutorial](https://howtodoinjava.com/java/date-time/format-time-12-hours-pattern)?

noelmathewisaac (commented):
Overall I think that the code largely adheres to the coding standards. Making the changes mentioned by abstracting out more functions and using a library for date/time conversions/formatting will improve the readability and maintainability of the code.

src/main/java/Parser.java
```diff
@@ -0,0 +1,231 @@
...
+     */
+    public void process() {
+        try {
+            if (command.equals("list")) {
+                ui.replace("Here are the tasks in your list:");
+                for (int i = 0; i < taskList.numOfTasks(); i++) {
+                    Integer taskNum = i + 1;
+                    ui.replace(taskNum.toString() + "." + taskList.getTask(i).toString());
+                }
+            } else if (command.contains("todo")) {
```
gycc7253: If i have a command
'event todo a task /by tmr'
this string contains 'todo' but it is actually event...
i think testing prefix maybe safer?

src/main/java/Storage.java
```diff
@@ -0,0 +1,101 @@
...
+                    for (int i = 0; i <= twbLastIndex; i++) {
+                        time += twb[i];
+                    }
+                    yearLine = inputArr[lastIndex - 1];
+                    String[] yearArr = yearLine.split("");
+                    for (int i = 0; i < 4; i++) {
+                        year += yearArr[i];
+                    }
+                    date = inputArr[lastIndex - 3] +
+                            " " + inputArr[lastIndex - 2] + " " + year;
```
gycc7253: i think over here, by coding standard, the + should be the start of the preceding line hh,
same mistake made bro.

src/test/java/TaskTest.java
```diff
@@ -0,0 +1,18 @@
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
gycc7253, line 3: could you please double check if static import needs to be imported before non-static, i think so but i am not 100% sure haha

src/main/java/Storage.java
```diff
@@ -0,0 +1,101 @@
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class Storage {
```
gycc7253: public class and methods undocumented bro, small issue, maybe u will do it later haha.

src/main/java/Event.java
```diff
@@ -0,0 +1,93 @@
+/*
+ * A type of Task.
+ */
+public class Event extends Task {
+    public Event(String description, String date, String time) {
```
gycc7253: i think public constructor needs a documentation also

src/main/java/Duke.java
```diff
@@ -1,10 +1,53 @@
+import java.util.Scanner;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.io.IOException;
```
gycc7253: maybe import of java.io should be in front of java.nio/util coz of lexicographic order

gycc7253 (commented):
haha nice reviewing your code, good work buddy!

src/main/java/Duke.java
cheunggalen: Will correct it, thanks.

src/main/java/Duke.java
cheunggalen: Thanks :)

src/main/java/Event.java
cheunggalen: Noted, thanks.

src/main/java/Event.java
cheunggalen: I will check and correct them, thanks.

src/main/java/Parser.java
cheunggalen: Noted, will change it, thanks.

src/main/java/Parser.java
cheunggalen: Ah yes, that will make it a lot easier, thanks!

src/main/java/Storage.java
cheunggalen: Will take note, thanks.

src/main/java/Storage.java
cheunggalen: Oh yes, thanks for pointing it out.

src/main/java/Task.java
cheunggalen: Noted, thanks for sharing the link!

src/main/java/Ui.java
cheunggalen: I will change it, thanks.

PR #246, opened 2021-01-31
[Glenn Lee] iP

glennljs:


src/main/java/checklst/task/Event.java
```diff
@@ -0,0 +1,58 @@
...
+import java.time.format.DateTimeParseException;
+
+import checklst.exception.ChecklstException;
+
+/**
+ * The Event class represents a Task which has a specific DateTime.
+ */
+public class Event extends Task {
+
+    protected final LocalDate dateTime;
```
zhoube: I would prefer to use `date` instead of `dateTime`, as dateTime suggests that the information also includes a time element

zhoube (commented):
Very well done! I believe you have adhered to almost all of the Java Coding Standards!

src/main/java/checklst/exception/ChecklstException.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Custom exception to represent Exceptions in the Checklst program.
+ */
+public class ChecklstException extends Exception {
+
+    /**
+     *
+     */
+    private static final long serialVersionUID = 4447981462872622989L;
```
yienyoong, line 11: Perhaps this should be 'SERIAL_VERSION_UID' as it is a constant?

src/main/java/checklst/task/Deadline.java
```diff
@@ -0,0 +1,58 @@
...
+public class Deadline extends Task {
+
+    protected final LocalDate dueDate;
+
+    protected Deadline(String name, LocalDate dueDate) {
+        super(name);
+        this.dueDate = dueDate;
+    }
+
+    protected Deadline(String name, boolean completed, LocalDate dueDate) {
```
yienyoong, line 21: Maybe boolean could be named isCompleted instead to sound more like a boolean?

src/main/java/checklst/parser/Parser.java
```diff
@@ -0,0 +1,105 @@
...
+            case "save":
+                storage.saveToFile();
+                ui.sendOutput("History sucessfully saved!");
+                break;
+            case "find":
+                ui.sendOutput("Here are the matching tasks in your list!\n\t" + taskList.findTask(input[1]));
+                break;
+            default:
+                throw new ChecklstException("Sorry I didn't understand that command!!");
+            }
```
yienyoong: I like that you followed the switch coding format! I sometimes forget to do this.

yienyoong (commented):
Your code is easy to read and largely follows coding standards. Nice work!

src/main/java/checklst/exception/ChecklstException.java
glennljs: I also thought so but apparently this is some optional requirement by java and is meant to be spelled like this? I'm not too sure because my IDE (vscode) added it in automatically.

src/main/java/checklst/task/Deadline.java
glennljs: Thanks for the suggestion! Changed accordingly.

src/main/java/checklst/parser/Parser.java
glennljs: Thanks!

src/main/java/checklst/task/Event.java
glennljs: Thanks for the suggestion! Changed accordingly.

kieron560 (commented):
Hi, I was supposed to review your code, but it is very clean and easy to understand! Good job!

PR #248, opened 2021-01-31
[Kuek Yan Ling] iP

yanlingkuek:


src/main/java/Duke.java
```diff
@@ -1,10 +1,189 @@
...
+
+        List<String> lst = new ArrayList<>();
+
+        while (!input.equals("bye")) {
+            if (!input.equals("list")) {
+                lst.add(input);
+                System.out.println("added: " + input);
+            } else {
+                for (int i = 1; i <= lst.size(); i++) {
+                    System.out.println(i + ". " + lst.get(i-1));
```
vvan-essa: Perhaps there should be whitespace on the left and right of the "-" sign? (: 
list.get(i - 1);

src/main/java/Duke.java
```diff
@@ -1,10 +1,189 @@
+import java.util.*;
```
vvan-essa: Perhaps imported classes should be listed explicitly? (:

src/main/java/Duke.java
```diff
@@ -1,10 +1,189 @@
...
+        while (!input.equals("bye")) {
+            System.out.println(input);
+            input = sc.nextLine();
+        }
+
+        System.out.println("Bye. Hope to see you again soon!");
+    }
+
+
+    static void level2() {
```
vvan-essa: Perhaps the method names could be more descriptive? In verb form? (:

src/main/java/Duke.java
```diff
@@ -1,10 +1,189 @@
...
+                }
+            }
+            input = sc.nextLine();
+        }
+
+        System.out.println("Bye. Hope to see you again soon!");
+
+    }
+
+
```
vvan-essa: Perhaps more javadoc comments for the methods? (:

src/main/java/DeadlineTask.java
```diff
@@ -0,0 +1,15 @@
+public class DeadlineTask extends Task {
+
+    protected String deadline;
+
+    public DeadlineTask(String description, String deadlline) {
```
markuz5116: Perhaps there is a typo for a variable name for deadline rather than deadlline.

src/main/java/Duke.java
markuz5116: importing tasks individually can be clearer.

src/main/java/Duke.java
```diff
@@ -1,10 +1,189 @@
+import java.util.*;
+
 public class Duke {
+
+    static void level1() {
```
markuz5116: Perhaps having a method name that briefly describes what it does would be clearer.

src/main/java/Duke.java
markuz5116: Not sure if I liked that the - does not have whitespace before and after it.

src/main/java/Duke.java
```diff
@@ -1,10 +1,189 @@
...
+            }
+            input = sc.nextLine();
+        }
+
+        System.out.println("Bye. Hope to see you again soon!");
+
+    }
+
+
+    static void level3() {
```
markuz5116: I would personally an access modifier for my method to show whether it can be used outside the package or class.

src/main/java/Duke.java
```diff
@@ -1,10 +1,189 @@
...
+        List<Task> lst = new ArrayList<>();
```
markuz5116: Perhaps instead of lst, using tasks could be clearer to know what the list contains.

src/main/java/Duke.java
```diff
@@ -1,10 +1,189 @@
...
+        lst.add(new Task("read book"));
+        lst.add(new Task("return book"));
+        lst.add(new Task("buy bread"));
+        lst.get(0).markAsDone();
+
+        Scanner sc = new Scanner(System.in);
+        String input = sc.nextLine();
+        String[] inputArray = input.split(" ");
```
markuz5116: perhaps rather than inputArray, you can have a plural form of the input to show that it has multiple parts.

src/main/java/Duke.java
```diff
@@ -1,10 +1,189 @@
...
+            input = sc.nextLine();
+            inputArray = input.split(" ");
+            if (inputArray[0].equals("list")) {
+                System.out.println("Here are the tasks in your list:");
+                for (int i = 1; i <= lst.size(); i++) {
+                    System.out.println(i + "." + lst.get(i - 1));
+                }
+            } else if (inputArray[0].equals("todo") || inputArray[0].equals("deadline") || inputArray[0].equals("event")) { //adding a task
+                Task task = new Task("");
+                String[] inputArr1;
```
markuz5116: perhaps extracting all the code that deals with different types of tasks could make your code clearer.

src/main/java/Task.java
```diff
@@ -0,0 +1,21 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
```
markuz5116: I like the name of this boolean, clearly tells me what it does.

markuz5116 (commented):
I personally think that access modifier should be included for all methods so that readers can know whether the method should be accessed from outside the class, can be clearer for the readers.

src/main/java/DeadlineTask.java
NiniJiaying: Perhaps there can be Javadoc comments to tell readers how this method works.

src/main/java/Duke.java
NiniJiaying: Perhaps it would be clearer if this method name starts with a verb?

src/main/java/Duke.java
```diff
@@ -1,10 +1,189 @@
...
+    }
+
+
+    static void level2() {
+        System.out.println("Hello! I'm Duke\n" + "What can I do for you?");
+
+        Scanner sc = new Scanner(System.in);
+        String input = sc.nextLine();
+
+        List<String> lst = new ArrayList<>();
```
NiniJiaying: Perhaps the parameter name can be something like commands? lst may sound a bit unclear to readers.

src/main/java/DeadlineTask.java
```diff
@@ -0,0 +1,15 @@
+public class DeadlineTask extends Task {
```
NiniJiaying: Maybe you can consider putting classes into packages.

PR #252, opened 2021-02-01
[Goh Wei Kiat] iP

mrweikiat:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
+public class Deadline extends Task {
+
+    protected String by;
```
nhzaci, line 3: Good use of access modifier, however, you could consider changing the variable name to a more descriptive one instead of 'by'

PR #258, opened 2021-02-02
[{Shi Zheng}] iP

Shizheng001:


src/main/java/Duke.java
```diff
@@ -1,10 +1,95 @@
...
+    }
+
+    /**
+     * prints Goodbye message.
+     */
+    public static void bye() {
+        System.out.println("Bye. Hope to see you again soon!");
+    }
+
+    public static void errorHandling(String input) throws DukeException {
```
BigDoot: Following the coding guidelines, method names should be verbs. Perhaps handleError() would be a better method name?

src/main/java/Task.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
+    }
+
+    public String getDescription() {
+        return this.description;
+    }
+
+    public Task changeStatus(Task t) {
```
BigDoot: Since this method is a setter, maybe setStatusTrue() would be better? It would explain to readers what status you are changing to as well!

BigDoot (commented):
There isn't very many issues with naming guidelines at the time of this review(as there aren't many methods yet), other than a few minor nitpicks! Fix them and you are good to go!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
+public class Deadline extends Task {
+
+    public Deadline(String description) {
+        super(description);
+    }
+
+    public Deadline(String description, boolean b) {
```
chewwh09: instead of using b, you could just use isDone.

src/main/java/Duke.java
```diff
@@ -1,10 +1,95 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+
 public class Duke {
+
+    /**
+     * main method to run the program
+     *
+     * @param args command line arguments taken in
+     */
```
Jonathan-Cao: Minor nitpick but perhaps consider starting Javadoc sentences with an uppercase letter and ending with a full stop to keep within the coding standard.

src/main/java/Duke.java
```diff
@@ -1,10 +1,95 @@
...
+
+                } else if (input.length() >= 8 && input.substring(0,8).equals("deadline")) {
+                    items.add(new Deadline(input.substring(8)));
+                } else if (input.length() >= 4 && input.substring(0,4).equals("todo")) {
+                    items.add(new Todo(input.substring(4)));
+                } else if (input.length() >= 5 && input.substring(0,5).equals("event")) {
+                    items.add(new Event(input.substring(5)));
+                } else if (input.length() >= 6 && input.substring(0,6).equals("delete")) {
+                    int taskToDelete = Integer.parseInt(input.substring(7));
+                    Task toRemove = items.get(taskToDelete - 1);
```
Jonathan-Cao: Perhaps int taskToDelete could be indexOfTaskToDelete even though it may be quite long.
It may help to distinguish it from Task toRemove.

src/main/java/Event.java
```diff
@@ -0,0 +1,14 @@
+public class Event extends Task {
+        public Event(String description) {
+            super(description);
+        }
+
+        public Event(String description, boolean b) {
+            super(description,b);
```
hengyiqun: Perhaps you could use isDone instead of b, so that it is more descriptive?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
...
+        super(description,b);
```
hengyiqun: For greater readability, perhaps you could have a whitespace after the comma?

src/main/java/Event.java
```diff
@@ -0,0 +1,14 @@
+public class Event extends Task {
+        public Event(String description) {
```
hengyiqun: Should there be an empty line before this constructor?
I notice that there are times you leave an empty line before constructors, and there are times you leave out the empty line.

src/main/java/Duke.java
hengyiqun: Same sentiment as the above!

hengyiqun (commented):
Clear code overall. Just a few nits to work out.

PR #262, opened 2021-02-03
[Wei Xue] iP

weixue123:


src/main/java/CustomExceptions/DateTimeFormatNotRecognizedException.java
```diff
@@ -0,0 +1,8 @@
+package CustomExceptions;
+
+public class DateTimeFormatNotRecognizedException extends Exception {
```
BenedictBCJJ: I'm not so sure for exceptions but class names should be nouns according to the coding standard.
Maybe DateTimeFormatException is better since it already implies that it is invalid/not recognised.

src/main/java/Parser.java
```diff
@@ -0,0 +1,152 @@
...
+import java.time.LocalTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.util.ArrayList;
+import java.util.Arrays;
+
+public class Parser {
+    private final String input;
+    private static final ArrayList<String> validActions
+            = new ArrayList<>(Arrays.asList("todo", "deadline", "event", "done", "delete", "list", "bye"));
```
BenedictBCJJ: Since this is a constant, it should follow the coding standard of all uppercase and using underscore to separate words.

src/main/java/Parser.java
```diff
@@ -0,0 +1,152 @@
...
+    private String getAtString() {
+        String action = this.getAction();
+        String remainingTokens = this.getRemainingTokens();
+        if (action.equals("event") && remainingTokens.contains("/at")) {
+            return remainingTokens.split("/at", 2)[1].trim();
+        }
+        return "";
+    }
+
+    public boolean inputsAreValid() {
```
BenedictBCJJ: Instead of naming inputsAreValid, areInputsValid seem more inline for a boolean return value method.

src/main/java/Tasks/Event.java
```diff
@@ -0,0 +1,21 @@
+package Tasks;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class Event extends Task {
+    private final LocalDateTime at;
```
BenedictBCJJ: I think 'at' is too short a name for a variable that has a rather large scope, should Event class become bigger in the future it might cause confusion in the future.
Since it might mean time or location.

BenedictBCJJ (commented):
Overall I think the project is very well segmented and can easily expand in scope in the coming weeks. Good job.

data/tasks.txt
```diff
@@ -0,0 +1,4 @@
+T | 1 | CS2107 Revision |
```
spzj: Hi Wei Xue! You can try to delete the output lines before committing, just in case it affects with testing.

src/main/java/Parser.java
```diff
@@ -0,0 +1,152 @@
+import CustomExceptions.*;
```
spzj: Perhaps you can list out the exceptions that you will be using in this file.

src/main/java/Ui.java
```diff
@@ -0,0 +1,106 @@
...
+                continue;
+            }
+
+            String action = parser.getAction();
+            String description = parser.getDescription();
+            LocalDateTime by = parser.getBy();
+            LocalDateTime at = parser.getAt();
+
+            switch (action) {
+                case "bye":
```
spzj: According to the modules' coding standards, case and default should be at the same indentation as the switch statement.

spzj (commented):
Overall, good job Wei Xue! Rmb to include javadocs too.

src/main/java/duke/tasks/Storage.java
```diff
@@ -0,0 +1,95 @@
...
+        String description = taskDetailsArray[2].trim();
+        String dateTimeString = taskDetailsArray[3].trim();
+        LocalDateTime dateTime = Parser.convertToDateTime(dateTimeString);
+
+        Task newTask;
+        if (taskType.equals("T")) {
+            newTask = new ToDo(description);
+        } else if (taskType.equals("D")) {
+            newTask = new Deadline(description, dateTime);
+        } else {
```
nicoleang09, line 115: Maybe instead of using the else block to handle the case of an Event task, it can be used to handle invalid task types?

src/main/java/duke/tasks/Storage.java
```diff
@@ -0,0 +1,95 @@
...
+        String dateTimeString = "";
+        String taskType;
+
+        if (task instanceof Deadline) {
+            taskType = "D";
+            dateTimeString = ((Deadline) task).getByDateTimeString();
+        } else if (task instanceof Event) {
+            taskType = "E";
+            dateTimeString = ((Event) task).getAtDateTimeString();
+        } else {
```
nicoleang09, line 147: Should the block to handle to do tasks be stated explicitly in an else-if block instead?

src/main/java/duke/ui/Parser.java
```diff
@@ -0,0 +1,172 @@
...
+
+    private static String getDescription(String input) {
+        String action = getAction(input);
+        String remainingTokens = getRemainingTokens(input);
+
+        if (action.equals("DEADLINE") && remainingTokens.contains("/by")) {
+            return remainingTokens.split("/by")[0].trim();
+        } else if (action.equals("EVENT") && remainingTokens.contains("/at")) {
+            return remainingTokens.split("/at")[0].trim();
+        } else {
```
nicoleang09, line 136: Same comment as before.

nicoleang09 (commented):
Your code quality is good and I could barely find any violations. I like how you broke bigger functions up into smaller ones and made your code easy to understand. Good job!

PR #271, opened 2021-02-06
Merge Level 9

KevinLohJunYong:


src/main/java/BadDateException.java
```diff
@@ -0,0 +1,6 @@
+public class BadDateException extends Exception {
+    @Override
+    public String getMessage() {
+        return "Please format your dead as yyyy-mm-dd";
```
danielonges, line 4: Might have meant `date` instead of `dead`?

src/main/java/Command.java
```diff
@@ -0,0 +1,16 @@
+public abstract class Command {
+    protected String line;
+    int itemNo = -1;
```
danielonges, line 3: Perhaps it would be better to just let the child classes that require this variable implement it instead? If not it would be slightly redundant for the other classes to contain it.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,32 @@
...
+
+    LocalDate localDate;
+
+    String formattedDescription;
+
+    String dateString;
+
+    Deadline(String description) throws EmptyArgumentException, BadDateException {
+        super(description);
+        int indexOfDate = description.indexOf("/") + 4;
```
danielonges, line 15: Might want to declare what the value `4` means? If not will be classified as magic number.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,32 @@
...
+        try {
+            localDate = LocalDate.parse(description.substring(indexOfDate));
+            dateString = localDate.format(DateTimeFormatter.ofPattern("MMM dd yyyy"));
+            formattedDescription = description.substring(0, indexOfDate) + dateString;
+        }
+        catch(DateTimeParseException e) {
```
danielonges, line 21: Can leave a space between `catch (`.

src/main/java/Command.java
```diff
@@ -0,0 +1,16 @@
...
+    public Command(String line) {
+        this.line = line;
+    }
+
+    public Command() {
+
+    }
+
+    public abstract boolean isExit();
+    public abstract String getType();
```
danielonges, line 13: Might want to consider implementing an enumeration for the type?

src/main/java/DeleteCommand.java
```diff
@@ -0,0 +1,19 @@
...
+        return false;
+    }
+    @Override
+    public String getType() {
+        return "delete";
+    }
+
+    @Override
+    public String getDescription() {
+        return null;
```
danielonges, line 17: Might not be a good idea to return null; could potentially result in NullPointerExceptions somewhere down the line.

src/main/java/Duke.java
```diff
@@ -1,10 +1,41 @@
...
+        } catch (IOException | EmptyArgumentException e) {
+            ui.showLoadingError();
+            tasks = new TaskList();
+        }
     }
-}
+
+    public void run() throws EmptyArgumentException, BadDateException, InvalidCommandException, IOException {
+        ui.sendWelcomeMessage();
+        while(true) {
```
danielonges, line 29: Might want to leave space between `while (true)`

src/main/java/Parser.java
```diff
@@ -0,0 +1,27 @@
+public class Parser {
+    public Command parseCommand(String line) throws InvalidCommandException {
+        String[] inputs = line.split(" ");
+        if(inputs[0].equals("list")) {
```
danielonges, line 4: Could consider using switch statement instead to make the code look cleaner!

src/main/java/Storage.java
```diff
@@ -0,0 +1,59 @@
...
+        FileWriter writer = new FileWriter(filePath);
+        for(Task task : tasks) {
+            writer.write(convertToFileFormat(task)+'\n');
+        }
+        writer.close();
+    }
+    private Task convertToTaskFormat(String line) throws EmptyArgumentException, BadDateException {
+        String[] arr = line.split(Pattern.quote(" | "));
+        Task task;
+        if(arr[0].charAt(0) == 'T') task = new ToDos(arr[2].trim());
```
danielonges, line 25: Might want to use curly braces instead of leaving inline if statements

src/main/java/TaskList.java
```diff
@@ -0,0 +1,47 @@
...
+        tasks = new ArrayList<>();
+    }
+    TaskList(List<Task> tasks) {
+        this.tasks = tasks;
+    }
+    public int numOfItems() {
+        return tasks.size();
+    }
+
+    public void run(Command command, Ui ui,Storage storage) throws EmptyArgumentException, BadDateException, InvalidCommandException, IOException {
```
danielonges, line 17: Might have exceeded line limit

src/main/java/TaskList.java
```diff
@@ -0,0 +1,47 @@
...
+        switch(command.getType()) {
+            case "deadline":
+                tasks.add(new Deadline(command.getDescription()));
+                ui.showUiForAdd(tasks.get(tasks.size()-1),numOfItems());
```
danielonges, line 21: Might want to include spacing between operators and operands

src/main/java/Ui.java
```diff
@@ -0,0 +1,52 @@
+import java.util.List;
+import java.util.List;
+import java.util.Scanner;
+
+public class Ui {
+    Scanner sc = new Scanner(System.in);
+    String logo =
+            " __        _        \n"
+                    + "|  _ \\ _   | | __ \n"
```
danielonges, line 9: Consider leaving 8 spaces of indentation for line breaks

danielonges (commented):
Might want to consider following the coding standard more, and cleaning up some of the logic of the code generally. Other than that, it was quite a pleasant experience reading your code! ☺️

damithc:
Looks like you PR'ed against the wrong repo? :-) Closing...

PR #276, opened 2021-02-06
[Jared Lim] iP

jared98lyj:


src/main/java/Duke.java
```diff
@@ -1,10 +1,135 @@
...
+/**
+ * Driver class that takes in input and performs certain functions based on input.
+ */
 public class Duke {
+
+    protected static ArrayList<Task> taskList = new ArrayList<>();
+
+    /**
+     * Main class to take in user input.
+     * @param args Filler
```
justgnohUG: Perhaps a more meaningful name for @param could be possibly replaced here

src/main/java/Event.java
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Main Constructor.
+     * @param description Description of given task.
+     */
+    Event(String description) {
+        super(description);
+    }
+
+    /**
+     * Overriden constructor to account for deadline and events that have timestamps.
```
justgnohUG: Not too sure if "Overriden" is the appropriate term here. I think "Overloaded constructor" may be more appropriate.

src/main/java/Duke.java
```diff
@@ -1,10 +1,135 @@
...
+    /**
+     * Adds current task to list of tasks.
+     * @param currentTask Current task.
+     */
+    protected static void addToTasks(Task currentTask) {
+        taskList.add(currentTask);
+    }
+
+    /**
+     * Enumerates all tasks in the lis using 1-based indexing.
```
justgnohUG: Minor typo for "list"

src/main/java/Task.java
```diff
@@ -0,0 +1,55 @@
...
+        this.isDone = false;
+        this.eventDate = "";
+    }
+
+    /**
+     * Overriden method to mark a task as done.
+     * @param description Description of method
+     * @param eventDate Boolean flag indicating whether task is done
+     */
+    public Task(String description, String eventDate) {
```
justgnohUG, line 40: Might be a good idea to add "@Override" tag if this method is truly overriden

src/main/java/Duke.java
```diff
@@ -1,10 +1,135 @@
+import java.util.ArrayList;
+import java.util.Scanner;
+
+/**
+ * Driver class that takes in input and performs certain functions based on input.
+ */
 public class Duke {
```
wei-yutong, line 9: this class is rather long, perhaps you could consider abstracting the methods out?

src/main/java/Duke.java
```diff
@@ -1,10 +1,135 @@
...
+        String command = "Hello";
+
+        while (!command.equalsIgnoreCase("bye")) {
+            if (command.equals("Hello")) {
+                System.out.println("Hello! I'm Duke");
+                System.out.println("What can I do for you?");
+            } else if (command.equals("list")) {
+                enumerateTasks();
+            } else if (command.startsWith("done")) {
+                String[] delString = command.split("\\s+");
```
wei-yutong: delString is not a very intuitive variable name, maybe you could make it more specific?

src/main/java/Event.java
wei-yutong: i think the term you meant to use might be "overloaded" instead

src/main/java/Task.java
```diff
@@ -0,0 +1,55 @@
+public class Task {
+
+    protected String description;
+    protected String eventDate;
+    protected boolean isDone;
+
+    enum Level {
+        low,
+        average,
+        high
```
wei-yutong: I think that constants in enum should be in all uppercase, eg "LOW", "AVERAGE" and "HIGH"

src/main/java/Task.java
```diff
@@ -0,0 +1,55 @@
...
+     * @param description Description of task
+     */
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+        this.eventDate = "";
+    }
+
+    /**
+     * Overriden method to mark a task as done.
```
wei-yutong: Did you mean "Overloaded" here as well? I've noticed it in a few other places. Otherwise, as the other reviewer mentioned, overridden methods should tagged with "@Override"

wei-yutong (commented):
your code is generally very clean and easy to read! i liked how your method names were clear and how you separated long lines accurately :)

PR #282, opened 2021-02-07
Level 10

ssagit:


src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,151 @@
...
+        return this.isBye;
+    }
+
+    /**
+     * Parses lines of user input and outputs corresponding command.
+     * @param tasks List of tasks from file.
+     * @param taskIterator Integer to count number of tasks at a time.
+     * @return New number of tasks after any addition or deletion.
+     */
+    public Tuple2<Integer, String> parseInput(String input, Task[] tasks, int taskIterator) {
```
car155, line 57: The parseInput method here is a bit long. Maybe its better to separate some of the parts like the deadline case into its own handleDeadline method for readability?

src/main/java/duke/ui/ConsoleUI.java
```diff
@@ -0,0 +1,132 @@
...
+                + "|____/ \\__,_|_|\\_\\___|\n";
+        System.out.println("Hello from\n" + logo);
+        String introduction = "I'm Duke!\nWhat can I do for ya?\n";
+        formatBox(introduction);
+    }
+
+    /**
+     * Duke says bye
+     */
+    public String bye() {
```
car155, line 40: Maybe use a verb for the method here? Something like sayBye()

src/main/java/duke/ui/ConsoleUI.java
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * Prints the tasks in the array of tasks
+     *
+     * @param taskArr array of Task objects
+     */
+    public String list(Task[] taskArr) {
+        String output = "++++++++++++++++++++++++++++++++++++\n" +
+                "Here are the tasks in your list: \n" +
+                "TaskType | isDone | taskName | time (if any)\n";
```
car155, line 105: Good indenting! I think it can be more readable with the plus in the front of each new line instead of at the back of the previous one.
```suggestion
        String output = "++++++++++++++++++++++++++++++++++++\n" 
                + "Here are the tasks in your list: \n" 
                + "TaskType | isDone | taskName | time (if any)\n";
```

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,151 @@
...
+    /**
+     * Exception class for unknown input parameters.
+     */
+    static class UnknownInputParamException extends Exception {
+        public UnknownInputParamException(String errorMessage) {
+            super(errorMessage);
+        }
+    }
+
+    public boolean getIsBye() {
```
creativequotient, line 47: Instead of `getIsBye()`, perhaps `isBye()` would be more intuitive and compact.

src/main/java/duke/taskclass/Task.java
```diff
@@ -0,0 +1,63 @@
...
+     * Function to mark a task as completed
+     */
+    public void markDone() {
+        this.isDone = true;
+    }
+
+    /**
+     * Returns a formatted string of the state of the task
+     */
+    public String toString() {
```
creativequotient, line 38: I think you can consider adding the `@Override` decorator here

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,113 @@
...
+     */
+    public Storage (String relativePath) {
+        this.relativePath = Paths.get(relativePath);
+        absolutePath = this.relativePath.toAbsolutePath();
+        taskText = new File(absolutePath.toString());
+
+        try {
+            if (!taskText.exists()) {
+                System.out.println("new file created");
+                taskText.createNewFile(); // creates the file if it doesn't exist
```
lrj689, line 38: The result of createNewFile() is unused, maybe try to put it with another if else statement

src/main/java/duke/parser/Parser.java
lrj689, line 47: Maybe just isBye() is enough

lrj689 (commented):
Good stuff. Keep it up!

damithc:
Looks like you PR'ed against the wrong repo? :-) Closing...

